PR #3, opened 2024-06-07
Add unit test for queue consumer

nerdai:


tests/message_queue_consumers/test_base.py
```diff
@@ -0,0 +1,40 @@
...
+
+@pytest.mark.asyncio()
+async def test_consumer_consumes_messages() -> None:
+    # Arrange
+    consumer_one = MockMessageConsumer()
+    mq = SimpleMessageQueue()
+    task = asyncio.create_task(mq.start())
+
+    # Act
+    await consumer_one.start_consuming(message_queue=mq)
```
nerdai, line 27: @logan-markewich thoughts on this?

I kinda feel like its redundant, but I know RabbitMQ has something like this where you define a consumer and you can start consuming from it as well.

I say its redundant cause you can do this from the message queue side using `mq.register_consumer()` (in fact that's what's being called here under the hood in `consumer.start_consuming(mq)`.

logan-markewich: Hmm yea, I guess is there a use case for having both directions (i.e. register vs explicitly calling start_consuming)?

nerdai: Keeping it for now then lol

PR #5, opened 2024-06-07
WIP init control plane

logan-markewich:
very WIP

Things to refactor
- make ActionType a string
- improve the pattern for setting up consumers
- the message queue is extremely coupled. Probably, we need a message queue client (which could be the message queue itself, or something that makes API calls?)

Some code to test this

```python
from agentfile.launchers.local import LocalLauncher
from agentfile.agent_server.fastapi import FastAPIAgentServer
from agentfile.control_plane.fastapi import FastAPIControlPlane
from agentfile.message_queues.simple import SimpleMessageQueue

from llama_index.core.agent import FunctionCallingAgentWorker
from llama_index.core.tools import FunctionTool
from llama_index.llms.openai import OpenAI

# create an agent
def get_the_secret_fact() -> str:
    """Returns the secret fact."""
    return "The secret fact is: A baby llama is called a 'Cria'."

tool = FunctionTool.from_defaults(fn=get_the_secret_fact)

worker = FunctionCallingAgentWorker.from_tools([tool], llm=OpenAI())
agent = worker.as_agent()

# create our multi-agent framework components
message_queue = SimpleMessageQueue()
control_plane = FastAPIControlPlane(message_queue=message_queue)
agent_server = FastAPIAgentServer(agent, message_queue)

# launch it
launcher = LocalLauncher([agent_server], control_plane, message_queue)
launcher.launch_single("What is the secret fact?")
```


agentfile/control_plane/base.py
nerdai, line 1: right, so this represents the top-level agent that does the coordination of agent consumers?

question: would it be possible for the agent servers to submit a message the queue themselves, or it has to go through the control_plane (as an exchange)?

nerdai (approved):
wooohooo!

agentfile/control_plane/base.py
logan-markewich: I think they could submit to the queue, if they were calling other agents as tools to complete their task 👀

PR #11, opened 2024-06-11
Logan/service base class

logan-markewich:
Bit of a refactor
- change a lot of language to revolve around "services" (Is this the right word??)
- introduced a BaseService class
- refactored the AgentService (previous FastAPIAgentServer) to have less scope (we can add more function later) and also implement the new base class
- other nits


agentfile/services/base.py
```diff
@@ -0,0 +1,54 @@
...
+
+from llama_index.core.bridge.pydantic import BaseModel
+
+from agentfile.messages.base import QueueMessage
+from agentfile.message_queues.base import BaseMessageQueue
+from agentfile.message_consumers.base import BaseMessageQueueConsumer
+from agentfile.types import ServiceDefinition
+
+
+class BaseService(ABC, BaseModel):
```
nerdai: Should we include PublisherMixin. 

Also wondering if we want to do as_publisher() like as_consumer() how that might look

nerdai (approved):
I think the change to "service" throughout is really good. It's more general which makes sense when we expand to various service types!

agentfile/services/base.py
logan-markewich: Ah good catch, lemme add that

PR #12, opened 2024-06-11
Add `QueueMessageStats` class and field to `QueueMessage`

nerdai:
For logging various desirable message epochs, introduce `QueueMessageStats`
- `publish_time`
- `start_processing_time`
- `end_processing_time`

### DIFF
--- a/agentfile/message_queues/simple.py
+++ b/agentfile/message_queues/simple.py
@@ -64,7 +64,9 @@ class SimpleMessageQueue(BaseMessageQueue):
                     message: QueueMessage = queue.popleft()
                     message.stats.process_start_time = message.stats.timestamp_str()
                     await self._publish_to_consumer(message)
-                    message.stats.process_end_time = message.stats.timestamp_str()
+                    message.stats.process_end_time = (
+                        message.stats.timestamp_str()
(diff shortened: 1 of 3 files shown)


agentfile/message_queues/simple.py
```diff
@@ -61,8 +61,10 @@ async def start(self) -> None:
         while self.running:
             for queue in self.queues.values():
                 if queue:
-                    message = queue.popleft()
+                    message: QueueMessage = queue.popleft()
+                    message.stats.process_start_time = message.stats.timestamp_str()
                     await self._publish_to_consumer(message)
+                    message.stats.process_end_time = message.stats.timestamp_str()
```
logan-markewich: So I think this works for now, but with actual servers running, publishing will be pretty fast (since it's just an api call to add it to the queue right?)

logan-markewich: I wonder if we need a more dedicated ack, so that we can better capture the processing time

nerdai: i'll leave a TODO note at least for it... i bet we'll run up into it once we move to actual server laucnher lol

PR #14, opened 2024-06-14
Add ToolService

nerdai:
- Endpoints:
    - `\` [GET]
    - `tool` [GET, POST]
    - `tool_call` [POST]
- Added `ToolCall` and `ToolCallResult` types
- Added  variants to `ActionTypes`: `NEW_TOOL_CALL` and `COMPLETED_TOOL_CALL`

### DIFF
--- a/agentfile/message_queues/simple.py
+++ b/agentfile/message_queues/simple.py
@@ -2,6 +2,7 @@
 
 import asyncio
 import random
+import logging
 
 from collections import deque
 from typing import Any, Dict, List
@@ -9,6 +10,10 @@ from llama_index.core.bridge.pydantic import Field
 from agentfile.message_queues.base import BaseMessageQueue
 from agentfile.messages.base import QueueMessage
 from agentfile.message_consumers.base import BaseMessageQueueConsumer
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG)
+logging.basicConfig(level=logging.DEBUG)
 
 
 class SimpleMessageQueue(BaseMessageQueue):
@@ -41,7 +46,12 @@ class SimpleMessageQueue(BaseMessageQueue):
         message_type_str = message.type
 
         if message_type_str not in self.consumers:
-            raise ValueError(f"No consumer for {message_type_str} has been registered.")
+            logger.debug(
+                f"Failed to publish message. No registered consumer '{message_type_str}'."
+            )
+            raise ValueError(
+                f"No consumer for '{message_type_str}' has been registered."
+            )
 
         if message_type_str not in self.queues:
             self.queues[message_type_str] = deque()
--- a/agentfile/services/tool.py
+++ b/agentfile/services/tool.py
@@ -111,7 +111,6 @@ class ToolService(BaseService):
                     *self._outstanding_tool_calls.values()
                 ]
             for tool_call in current_tool_calls:
-                print(f"processing {tool_call}", flush=True)
                 tool = get_function_by_name(
                     self.tools, tool_call.tool_call_bundle.tool_name
                 )
--- a/agentfile/services/tool.py
+++ b/agentfile/services/tool.py
@@ -1,0 +1,180 @@
+import asyncio
+import uuid
+from contextlib import asynccontextmanager
+from fastapi import FastAPI
+from typing import Any, AsyncGenerator, Dict, List, Optional
+
+from llama_index.core.agent.function_calling.step import (
+    build_missing_tool_output,
+    get_function_by_name,
+)
+from llama_index.core.bridge.pydantic import PrivateAttr
+from llama_index.core.llms import ChatMessage, MessageRole
+from llama_index.core.tools import BaseTool, ToolSelection
+from llama_index.core.tools.calling import (
+    call_tool_with_selection,
+)
+
+from agentfile.message_consumers.base import BaseMessageQueueConsumer
+from agentfile.message_consumers.callable import CallableMessageConsumer
+from agentfile.message_publishers.publisher import PublishCallback
+from agentfile.message_queues.base import BaseMessageQueue
+from agentfile.messages.base import QueueMessage
+from agentfile.services.base import BaseService
+from agentfile.types import (
+    ActionTypes,
+    ToolCall,
+    ToolCallResult,
+    ServiceDefinition,
+    CONTROL_PLANE_NAME,
+)
+
+
+class ToolService(BaseService):
+    service_name: str
+    tools: List[BaseTool]
+    description: str = "Local Tool Service."
+    running: bool = True
+    step_interval: float = 0.1
+
+    _outstanding_tool_calls: List[ToolCall] = PrivateAttr()
+    _message_queue: BaseMessageQueue = PrivateAttr()
+    _app: FastAPI = PrivateAttr()
+    _publisher_id: str = PrivateAttr()
+    _publish_callback: Optional[PublishCallback] = PrivateAttr()
+
+    def __init__(
+        self,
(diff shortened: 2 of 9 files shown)


agentfile/services/tool.py
```diff
@@ -0,0 +1,180 @@
...
+                continue
+
+            current_tool_calls = self._outstanding_tool_calls
+            for tool_call in current_tool_calls:
+                tool = get_function_by_name(
+                    self.tools, tool_call.tool_selection.tool_name
+                )
+
+                tool_output = (
+                    call_tool_with_selection(tool_call.tool_selection, self.tools)
```
logan-markewich: Nit: let's use async `acall_tool_with_selection`

We could even asyncio.gather all tool calls 👀

logan-markewich: (Although tbh, most tools are sync anyways)

agentfile/services/tool.py
```diff
@@ -0,0 +1,180 @@
...
+    def publish_callback(self) -> Optional[PublishCallback]:
+        return self._publish_callback
+
+    async def processing_loop(self) -> None:
+        while True:
+            if not self.running:
+                await asyncio.sleep(self.step_interval)
+                continue
+
+            current_tool_calls = self._outstanding_tool_calls
```
logan-markewich: To make this slightly more async safe, we could copy the array rather than referencing it 

Something like 
`current_tool_calls = [*self._outstanding_tool_calls]`

agentfile/services/tool.py
```diff
@@ -0,0 +1,180 @@
...
+                        ).dict(),
+                    )
+                )
+
+            await asyncio.sleep(self.step_interval)
+
+    async def process_message(self, message: QueueMessage, **kwargs: Any) -> None:
+        if message.action == ActionTypes.NEW_TOOL_CALL:
+            tool_selection = ToolSelection(**message.data or {})
+            self._outstanding_tool_calls.append(tool_selection)  # is this async safe?
```
logan-markewich: Good note in the comment. Could be a lock 👀

agentfile/services/tool.py
nerdai: sounds good!

agentfile/services/tool.py
nerdai: yeah i think we'll need a lock to be safe here

agentfile/services/tool.py
nerdai: ohh interesting how unpacking the array would make it async safe! I had no idea...

I guess I would've just used the lock again here, to ensure I was getting the most up to date read of it.

tests/test_tool_service.py
```diff
@@ -0,0 +1,161 @@
...
+    # assert
+    assert server.tools == tools
+    assert multiply_tool_metadata == tools[0].metadata
+    assert server.running is False
+    assert server.description == "Test Tool Server"
+    assert server.step_interval == 0.5
+
+
+@pytest.mark.asyncio()
+async def test_create_tool_call(tools: List[BaseTool], tool_call: ToolCall) -> None:
```
nerdai, line 70: @logan-markewich: please let me know what you think

nerdai: UPDATE: sorry this is the wrong one!

tests/test_meta_service_tool.py
```diff
@@ -0,0 +1,214 @@
...
+    from_tool_service_kwargs: Dict[str, Any],
+) -> None:
+    # arrange
+    # act/assert
+    with pytest.raises(ValueError):
+        await MetaServiceTool.from_tool_service(**from_tool_service_kwargs)
+
+
+@pytest.mark.asyncio()
+async def test_tool_call_output(
```
nerdai, line 133: @logan-markewich: This is the one I wanted to flag you on...

agentfile/message_queues/simple.py
```diff
@@ -41,7 +46,12 @@ async def _publish(self, message: QueueMessage, **kwargs: Any) -> Any:
         message_type_str = message.type
 
         if message_type_str not in self.consumers:
-            raise ValueError(f"No consumer for {message_type_str} has been registered.")
+            logger.debug(
```
logan-markewich, line 49: always good to start a good habit of debug messages in a new repo 😀

agentfile/services/tool.py
```diff
@@ -0,0 +1,200 @@
...
+            async with self.lock:
+                current_tool_calls: List[ToolCall] = [
+                    *self._outstanding_tool_calls.values()
+                ]
+            for tool_call in current_tool_calls:
+                tool = get_function_by_name(
+                    self.tools, tool_call.tool_call_bundle.tool_name
+                )
+
+                tool_output = tool(
```
logan-markewich: should use async tool calls here?

agentfile/tools/meta_service_tool.py
```diff
@@ -0,0 +1,232 @@
...
+        # TODO by requests
+        # make a http request, try to parse into BaseTool
+        elif (
+            tool_service_url is not None
+            and tool_service_api_key is not None
+            and tool_service_name is not None
+        ):
+            return cls(
+                tool_service_name=tool_service_name,
+                tool_metadata=ToolMetadata("TODO"),
```
logan-markewich, line 110: heh -- yea, we can add URL requests in the future

tests/test_meta_service_tool.py
```diff
@@ -0,0 +1,214 @@
...
+    mq_task = asyncio.create_task(message_queue.start())
+    ts_task = asyncio.create_task(tool_service.processing_loop())
+
+    # act
+    tool_output = await meta_service_tool.acall(a=1, b=9)
+
+    # clean-up/shutdown
+    await asyncio.sleep(0.5)
+    mq_task.cancel()
+    ts_task.cancel()
```
logan-markewich, line 151: oh this is much nicer, I should add this to the local launcher

tests/test_meta_service_tool.py
logan-markewich: I think this test looks good! Was there something specific to look into here?

logan-markewich (approved):
This looks super good!

Do you have an example of the end-to-end UX for this? I wonder if we need to update the launcher to handle some of the tedious steps

tests/test_meta_service_tool.py
nerdai: oh sorry nothing specific, just so that it may give you context in case needed as you review rest of code!

tests/test_meta_service_tool.py
nerdai: 🙏

agentfile/services/tool.py
nerdai: Yeah I think so. Was going to ask you about that.

Currently `tool: BaseTool`,  but is it a seamless swap to just use `tool: AsyncBaseTool` instead?

agentfile/tools/meta_service_tool.py
nerdai: 👍 lol in time... probably once we have the local server launchers up and running we can actually do this.

agentfile/services/tool.py
logan-markewich: should be seamless! We also have a utility to adapt any base tool to an async tool

nerdai:
> This looks super good!
> 
> Do you have an example of the end-to-end UX for this? I wonder if we need to update the launcher to handle some of the tedious steps

Thanks for looking! I'm going to cook up an end-to-end UX for it, though I was thinking of just building on top of what you had with launcher and control plane. I was wondering if we should have users query the control plane to list the available tools in the tool service(s).

PR #18, opened 2024-06-20
Add example script for ToolService with Launcher

nerdai:


agentfile/orchestrators/agent.py
```diff
@@ -44,14 +44,12 @@ async def get_next_messages(
...
                 tools,
                 user_msg=task_def.input,
-                error_on_no_tool_call=False,
             )
         else:
             messages = memory.get()
             response = await self.llm.apredict_and_call(
                 tools_plus_human,
                 chat_history=messages,
-                error_on_no_tool_call=False,
```
logan-markewich: we should keep this line 👀 But just need the latest version of core (I updated the toml in main)

agentfile/launchers/local.py
```diff
@@ -87,14 +101,11 @@ async def alaunch_single(self, initial_task: str) -> None:
...
-
         # runs until the message queue is stopped by the human consumer
-        await self.message_queue.start()
+        mq_task = asyncio.create_task(self.message_queue.start())
+        await asyncio.sleep(10)
+
+        # shutdown
+        for task in bg_tasks:
+            task.cancel()
+        mq_task.cancel()
```
logan-markewich: I wonder if there's a way to return the final result here too -- maybe a future improvement

agentfile/orchestrators/agent.py
nerdai: Yeah i was confused why that was there, but my core wasn't updated on my local machine! Will make the change 👍🏼

agentfile/launchers/local.py
```diff
@@ -31,10 +32,12 @@ def __init__(
         services: List[BaseService],
         control_plane: BaseControlPlane,
         message_queue: SimpleMessageQueue,
+        additional_consumers: List[BaseMessageQueueConsumer] = [],
```
logan-markewich: nit: we should avoid having lists as default params, due to pass-by-reference

agentfile/launchers/local.py
nerdai: I modified this to what I think is a bit more graceful shutdown.

agentfile/launchers/local.py
nerdai: good catch! I removed this entirely anyways though.

PR #20, opened 2024-06-20
Human Service

nerdai:
- Add HumanService
- Modifies Agent Orchestrator to route to HumanService
- Add unit tests for HumanService
- Add example_script for HumanService (sort of works, needs polishing IMO)


agentfile/orchestrators/agent.py
```diff
@@ -73,13 +76,17 @@ async def get_next_messages(
                 name = source.tool_name
                 input_data = source.raw_input
                 input_str = next(iter(input_data.values()))
+                if name == "default_human_service":
+                    action = ActionTypes.REQUEST_FOR_HELP
```
nerdai: So, if the agent orchestrator determines that the `human_service` should be selected, which with this implementation should actually mean that human-service has a default name like CONTROL_PLANE_NAME.

Blanking on how else to make this more flexible. I mean we could tease out the human-service from the other services and have a special attribute/property of the Orchestrator?

agentfile/services/human.py
```diff
@@ -0,0 +1,209 @@
...
+                # create history
+                history = [
+                    ChatMessage(
+                        role=MessageRole.ASSISTANT,
+                        content=HELP_REQUEST_TEMPLATE_STR.format(
+                            input_str=task_def.input
+                        ),
+                    ),
+                    ChatMessage(role=MessageRole.USER, content=result),
+                ]
```
nerdai, line 155: I make a fictitious history here, as I know with the agent orchestrator it takes this and summarizes it to be the actual content of the assistant. Does that mean, we don't have to be so careful to adhere to the ordering of messages (i.e., roles) in creating this history?

agentfile/services/human.py
```diff
@@ -0,0 +1,209 @@
...
+                async with self.lock:
+                    del self._outstanding_human_requests[task_def.task_id]
+
+            await asyncio.sleep(self.step_interval)
+
+    async def process_message(self, message: QueueMessage, **kwargs: Any) -> None:
+        if message.action == ActionTypes.REQUEST_FOR_HELP:
+            task_def = TaskDefinition(**message.data or {})
+            async with self.lock:
+                self._outstanding_human_requests.update({task_def.task_id: task_def})
```
nerdai: I have a concern here which I don't know if is warranted around using `task_id` as the key. Is it possible for the agent orchestrator to send multiple `TaskDefinition`'s with same task_id to this human service? When I looked the orchestration code, it seemed that there was nothing preventing this from happening (I don't think its a big deal with the agent services, because they use `create_task(....)`).

example_scripts/human_local_service.py
```diff
@@ -0,0 +1,47 @@
...
+    orchestrator=AgentOrchestrator(llm=OpenAI()),
+)
+
+# launch it
+launcher = LocalLauncher(
+    [agent_service, human_service],
+    control_plane,
+    message_queue,
+)
+launcher.launch_single("What is 5 + 5?")
```
nerdai: This is kind of funny sometimes. A few times I'll supply the correct answer 10, but the agent orchestrator for some reason still takes the action of sending it back to the human service for another go round at it. Probably has happened 10-20% of the time that I've tested.

We may want to implement more strict orchestration logic after human intervention?

agentfile/orchestrators/agent.py
logan-markewich: I think the most generalized way would be to treat the human service the same as any other service? i.e. it gets NEW_TASK messages, returns COMPLETED_TASK messages

This works too though 👀

agentfile/services/human.py
logan-markewich: Yea the ordering of roles here is less important. I think this looks good!

agentfile/services/human.py
logan-markewich: hmm thats a good point actually. Agents use `create_task()` but they also set the `task_id` to match the `TaskDefinition()` 🤔 But, I think that would just overwrite the previous if it got sent twice, which is probably fine?

So TLDR, the orchestrator could send the same `task_id` twice, but if it just overwrites the previous in the service, that should be fine?

example_scripts/human_local_service.py
logan-markewich: I think this is even the case for agent services. I suspect the agent orchestrator needs some prompt engineering to make this slightly more reliable?

agentfile/services/human.py
nerdai: great! thanks for confirming

agentfile/services/human.py
nerdai: ah okay, yeah i think it would be fine for agent service, but that seems to be problematic for my current impl of human service.

to be safe, I'll track `_outstanding_human_requests` with a different id via some sort of lightweight container around the TaskDefinition

example_scripts/human_local_service.py
nerdai: Ah okay, so some improvements to be had here generally (future PR stuff, I guess). 👍

PR #23, opened 2024-06-21
[wip] server launcher + ui

logan-markewich:


agentfile/services/agent.py
```diff
@@ -239,5 +260,14 @@ async def reset_agent(self) -> Dict[str, str]:
...
         return {"message": "Agent reset"}
 
-    def launch_server(self) -> None:
-        uvicorn.run(self._app, host=self.host, port=self.port)
+    async def launch_server(self) -> None:
+        logger.info(f"Launching {self.service_name} server at {self.host}:{self.port}")
+        # uvicorn.run(self._app, host=self.host, port=self.port)
+
+        class CustomServer(uvicorn.Server):
+            def install_signal_handlers(self) -> None:
```
nerdai, line 268: Guessing this was causing us some issues? So we needed to override it?

agentfile/services/human.py
```diff
@@ -109,31 +131,18 @@ def publish_callback(self) -> Optional[PublishCallback]:
...
                 continue
 
             async with self.lock:
-                current_human_tasks = [*self._outstanding_human_tasks.values()]
-            for human_task in current_human_tasks:
-                task_def = human_task.task_definition
+                current_human_tasks = [*self._outstanding_human_tasks]
+
+            while len(current_human_tasks) > 0:
+                task_def = current_human_tasks.pop(0)
```
nerdai: 👏

agentfile/services/human.py
```diff
@@ -201,29 +205,76 @@ async def launch_local(self) -> None:
 
     # ---- Server based methods ----
 
-    @asynccontextmanager
-    async def lifespan(self) -> AsyncGenerator[None, None]:
-        """Starts the processing loop when the fastapi app starts."""
-        asyncio.create_task(self.processing_loop())
-        yield
-        self.running = False
```
nerdai, line 209: How come we don't need lifespan for human service? I see that we also added handle_task endpoint/endpoint-function that is different from the other services as well.

nerdai: I guess this means that as server, we want this service to handle task on request (versus listening to the message queue?)

nerdai (commented):
Neat to see textual -- it looks very intuitive!

agentfile/services/agent.py
logan-markewich: Yea, I needed some way to spin up multiple servers, and apparently this is the only way 😅

agentfile/services/human.py
logan-markewich: Yea if you think about it as a UI, we don't need any processing loop, it should be fully driven by the api endpoints. 

So, since there's no processing loop, I removed the lifespan

logan-markewich: (Let me know if I'm off base here, but it seems to work!)

logan-markewich:
Oops, did I botch tests lol

nerdai:
> Oops, did I botch tests lol

Haha all good. I can take a look after a quick bite ~15 mins.

nerdai (approved):

PR #31, opened 2024-06-22
[WIP] RemoteMessageQueueClient

nerdai:


agentfile/message_consumers/remote.py
```diff
@@ -12,7 +12,7 @@ class RemoteMessageConsumerDef(BaseModel):
     message_type: str = Field(
         default="default", description="Type of the message to consume."
     )
-    url: Optional[str] = Field(default=None, description="URL to send messages to.")
+    url: str = Field(default_factory=str, description="URL to send messages to.")
```
nerdai, line 15: Had to make this change cause when calling the `register_consumer` endpoint of message queue, the `RemoteConsumerDef` has a strict requirement of str (i.e. not Optional)

agentfile/message_queues/simple.py
```diff
@@ -72,13 +72,20 @@ def __init__(
             tags=["Consumers"],
         )
 
+        self._app.add_api_route(
+            "/publish",
+            self._publish,
```
nerdai, line 77: Note: the endpoint function here is `_publish` rather than `publish` because the former causes issues due to the `PublishCallback` param.

A potential resolution would be to make the PublishCallback an instance attribute of the class so that the `publish` function can serves as the endpoint function. A TODO for a future PR?

agentfile/message_consumers/remote.py
logan-markewich: Makes sense!

agentfile/message_queues/simple.py
logan-markewich: Yea I noticed this too. Let's clean it up later I think

logan-markewich (approved):

PR #32, opened 2024-06-23
Unit tests for remote client mq

nerdai:


llama_agents/message_queues/remote_client.py
nerdai, line 1: @logan-markewich: I am beginning to think that this remote client is not really general. Specifically, this client works more or less exclusively with SimpleMessageQueue.

As such, I am thinking to rename it to `SimpleRemoteClientMessageQueue` and actually place it in `simple.py`

logan-markewich: Yea I kind of agree -- its pretty specific to the endpoints declared in the simple message queue

llama_agents/message_queues/remote_client.py
```diff
@@ -49,18 +62,31 @@ async def deregister_consumer(
...
+        return result
 
     async def get_consumers(
         self, message_type: str, get_consumers_url: str = "get_consumers"
     ) -> List[BaseMessageQueueConsumer]:
         client_kwargs = self.client_kwargs or {}
         client = self.client or httpx.AsyncClient(**client_kwargs)
-        url = urljoin(self.base_url, get_consumers_url)
+        url = urljoin(self.base_url, f"{get_consumers_url}/{message_type}")
+        logger.info(f"url: {url}")
```
logan-markewich: This log probably needs a touch more info

nerdai: oh oops, that was mainly for my debugging during testing. I'm just going to remove this!

logan-markewich (approved):

llama_agents/message_queues/simple.py
```diff
@@ -66,7 +66,7 @@ def __init__(
         )
 
         self._app.add_api_route(
-            "/get_consumers",
+            "/get_consumers/{message_type}",
```
nerdai, line 155: @logan-markewich: changed this to take on path params as opposed to query params

llama_agents/message_queues/remote_client.py
```diff
@@ -6,17 +6,22 @@
...
+from llama_agents.types import PydanticValidatedUrl
 
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.DEBUG)
 logging.basicConfig(level=logging.DEBUG)
 
 
 class RemoteClientMessageQueue(BaseMessageQueue):
-    base_url: str
+    base_url: PydanticValidatedUrl
```
nerdai: @logan-markewich: while still a str type, this has to be a valid http or https url now

llama_agents/message_queues/simple.py
logan-markewich: I don't think I was using this API yet, but makes sense!

llama_agents/message_queues/remote_client.py
nerdai: Great, in that case I'll go ahead with the rename and moving to `message_queues.simple` module.

llama_agents/message_queues/remote_client.py
logan-markewich: There's a lot of places where I hardcoded `http://....` when creating URLs -- TODO: probably need a better pattern for that to account for https

PR #34, opened 2024-06-23
[WIP] add query rewriting notebook example 

jerryjliu:
WIP because the descirptions on the notebook need to be better.

setup two agents, a hyde agent and a simple agent, where hyde --> simple agent, combined using pipeline orchestrator.

depends on this PR being landed https://github.com/run-llama/llama_index/pull/14329


logan-markewich:
Very cool! Will make a release on llama-index so we can land this

logan-markewich:
@jerryjliu This should be good to go! Greatly expanded on the notebook

Issue #47, opened 2024-06-26
Add a k8s launcher


logan-markewich:
Assuming #46 is completed, we should be able to extend that into a fully scalable k8s deployment!

k8s can be confusing, especially for newcomers, so giving some way to help out with that is extremely valuable

nerdai:
I believe this should've been closed along with #66

PR #66, opened 2024-06-26
Launch with/without Docker as well as K8s Example

### LINKED ISSUE #46: Add a docker launcher
We should be able to take a system, and generate a docker deployment with some given specifications.

This would be a large undertaking! But would go a long way to making `llama-agents` more production ready

nerdai:
This PR adds an example of a multi-agent system build that you can launch via three methods:
- ServerLauncher (Without Docker)
- With Docker
- Kubernetes

closes #46 #47


.pre-commit-config.yaml
```diff
@@ -11,6 +11,7 @@ repos:
       - id: check-symlinks
       - id: check-toml
       - id: check-yaml
+        args: [--allow-multiple-documents]
```
nerdai, line 14: otherwise, it this hook complains about my k8s manifest files that contains multiple yaml documents in a single file.

logan-markewich (approved):

PR #79, opened 2024-06-29
[feature] Add query component service

jerryjliu:
- Allow any query component from llamaindex core be represented as a service in llama-agents
- This allows you to define more expressive query pipelines with the pipeline orchestrator (similar to how you would for query pipelines locally) 
- Expand ServiceComponent to better reflect the set of input keys you can pass in 
- Rewrote the Query Rewrite RAG notebook (and fixed the HyDE step)


llama_agents/orchestrators/pipeline.py
```diff
@@ -16,6 +17,82 @@
 RESULT_KEY = "result"
 
 
+INPUT_DICT_KEY = "__input_dict__"
+
+
+def get_service_component_message(
```
logan-markewich, line 23: Is there a specific reason to define these two classes outside of the orchestrator class?

jerryjliu: these two functions? 

they didn't seem to rely on `self` too much, but don't have a strong pref either way

jerryjliu:
@logan-markewich do you know why tests are failing here? they're passing locally

logan-markewich:
@jerryjliu 

```
ModuleNotFoundError: No module named 'llama_agents.services.component
```
I think you missed committing a file

jerryjliu:
> @jerryjliu
> 
> ```
> ModuleNotFoundError: No module named 'llama_agents.services.component
> ```
> 
> I think you missed committing a file

lol oops fixed

examples/query_rewrite_rag.ipynb
logan-markewich, line 1: oh my, this flow is... fun 😅 

I agree, we definitely need a more direct "here's a llama-index thing, here's your llama-agents thing" flow. This is ok-ish for now.

logan-markewich (approved):

PR #81, opened 2024-06-30
[example] reflection demo

jerryjliu:
~~WIP because notebook doesn't work - running into `ConnectError`. However the server/client python files work.~~

Created a simple demo to spin up self-reflection/critic agents as a service

relies on fix on introspective agents in main repo: https://github.com/run-llama/llama_index/pull/14474


examples/reflection/toxicity_reflection_service.ipynb
```diff
@@ -0,0 +1,790 @@
+{
```
nerdai, line 1: <p>Maybe at the end of this NOTE, we can add "(see <a href="see [here](https://docs.llamaindex.ai/en/latest/module_guides/deploying/agents/usage_pattern/#customizing-your-agent" target="_blank">here</a> for more details)", where here links to example usage of <code>FnAgentWorker</code> from our docs.</p><p> </p><p><strong>UPDATE.</strong> Nevermind...I read the next cell below this one which contains an example code snippet for <code>FnAgentWorker</code>.👍  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama-agents/pull/81/discussion/#comment-1660937766'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

nerdai (approved):
Nice to see this example being projected into llama-agents. There's def a few more steps involved in setting up a multi-agent system.

I wonder if we want to conform to a "standard" format in building these demos/cookbooks in the future:

1. Build Agent or Other Service(s), show they work as a standalone service by sending it tasks directly (maybe for only 1-2 of them)

2. Then say, "now that we know they work standalone ... let's connect it to a multi-agent system...". This is where we bring in the: message-queue, control-plane (and its orchestrator), and launcher.

It may become a bit repetitive, but perhaps that will be good for our users?

jerryjliu:
> Nice to see this example being projected into llama-agents. There's def a few more steps involved in setting up a multi-agent system.
> 
> I wonder if we want to conform to a "standard" format in building these demos/cookbooks in the future:
> 
> 1. Build Agent or Other Service(s), show they work as a standalone service by sending it tasks directly (maybe for only 1-2 of them)
> 2. Then say, "now that we know they work standalone ... let's connect it to a multi-agent system...". This is where we bring in the: message-queue, control-plane (and its orchestrator), and launcher.
> 
> It may become a bit repetitive, but perhaps that will be good for our users?

yeah that makes sense. see the CRAG PR https://github.com/run-llama/llama-agents/pull/82 for an example of this 

i do think in terms of future improvements:
1. we want to reduce the number of lines of code to convert agents into services
2. we may want to think about a "converter" function that can take in an existing query pipeline and auto-convert it into a set of services. that way the user can define the e2e flow locally, and then not have to worry about writing too much code to convert

PR #85, opened 2024-07-01
[Feature] Add RabbitMQMessageQueue (RabbitMQ Integration)

### LINKED ISSUE #25: Add RabbitMQ as a message queue
RabbitMQ is a popular message queue. For deployments and scalability, supporting it would be a huge win.

This may require tweaks to the message queue/publishing/consumer abstractions, but hopefully it fits nicely!

nerdai:
This PR adds RabbitMQMessageQueue (naming might be redundant?)

Highlights:
- Can use `RabbitMQMessageQueue` in place of `SimpleMessageQueue` in both `LocalLauncher` and `ServerLauncher`
     - The new `examples/rabbitmq` subdirectory contains installation and usage instructions as well as example scripts using both local and server launcher
     - Once this PR lands: will create a docker/k8s example using RabbitMQ
- Uses the `aio-pika` client for RabbitMQ which is async first
- Changes to `BaseMessageQueue` interface:
    - `register_consumer()` now returns a `StartConsumingCallable`
    - The consumer should invoke this function to start consuming from the queue


examples/rabbitmq/local_launcher_example.py
```diff
@@ -0,0 +1,72 @@
...
+# create an agent
+def get_the_secret_fact() -> str:
+    """Returns the secret fact."""
+    return "The secret fact is: A baby llama is called a 'Cria'."
+
+
+tool = FunctionTool.from_defaults(fn=get_the_secret_fact)
+
+worker1 = FunctionCallingAgentWorker.from_tools([tool], llm=OpenAI())
+# worker2 = FunctionCallingAgentWorker.from_tools([], llm=OpenAI())
```
logan-markewich: Remove this line?

examples/rabbitmq/sever_launcher_example.py
logan-markewich, line 1: heh good call on the examples folders, just noticed this. Probably other examples could be a bit more organized overall

examples/rabbitmq/local_launcher_example.py
nerdai: you bet!

llama_agents/message_queues/rabbitmq.py
```diff
@@ -0,0 +1,186 @@
...
+        password       = *( unreserved / pct-encoded / sub-delims )
+        vhost          = segment
+
+    The Work Queue created has the following properties:
+        - Exchange with name self.exchange
+        - Messages are published to this queue through the exchange
+        - Consumers are bound to the exchange and have queues based on their
+            message type
+        - Round-robin dispatching: with multiple consumers listening to the same
+            queue, only one consumer will be chosen dictated by sequence.
```
logan-markewich, line 59: Can we add a small `Args:` and `Examples:` section?

llama_agents/message_queues/rabbitmq.py
```diff
@@ -0,0 +1,186 @@
...
+    @classmethod
+    def from_url_params(
+        cls,
+        username: str,
+        password: str,
+        host: str,
+        port: Optional[int] = None,
+        secure: bool = False,
+        exchange_name: str = DEFAULT_EXCHANGE_NAME,
+    ) -> "RabbitMQMessageQueue":
```
logan-markewich, line 93: Needs a short docstring (doesn't have to list out args if you don't want, but just so that it shows up in api reference)

logan-markewich: (Similar comment for other non-dunder methods)

nerdai: Cool beans! I think I can follow format of doc strings used in `SimpleMessageQueue`?

llama_agents/message_queues/rabbitmq.py
nerdai: you bet!

logan-markewich (approved):
Looks pretty good to me!

Can you take a second pass through the README.md as well, and make sure nothing in there is outdated yet?

llama_agents/message_queues/rabbitmq.py
logan-markewich: Yea that works!

logan-markewich: (GitHub-copilot wrote most of these for me lol with some minor tweaks)

nerdai:
> Looks pretty good to me!
> 
> Can you take a second pass through the README.md as well, and make sure nothing in there is outdated yet?

Absolutely! Thanks for the turn around time on the review!

llama_agents/message_queues/rabbitmq.py
nerdai: man... i haven't taken the plunge yet to work with copilot....

nerdai:
> Looks pretty good to me!
> 
> Can you take a second pass through the README.md as well, and make sure nothing in there is outdated yet?

Okay, I updated it. The main update was that `mq.register_consumer()` returns a `StartConsumingCallable` which is async and that needs to be called in order for the consumer to actually start consuming (well not really for SimpleMessageQueue since its a dummy callable thats returned, but we should keep the pattern consistent)

PR #90, opened 2024-07-02
[BUG] Add README reminder to example users to create poetry.lock before first run

emekdahl:
Observed:

When following the docker k8s example, `docker-compose` command fails unless we first generate the `poetry.lock` file. Adding a friendly reminder to others following the example that they need to generate the poetry lock file on the first run.


logan-markewich (approved):
Thanks @emekdahl

Issue #97, opened 2024-07-03
Need to add execution delay in test client


peteryxu:
Following the docker sample readme.md, (https://github.com/peteryxu/llama-agents/tree/learn/examples/docker-kubernetes, ) running the Client test code, and got error below.  

Traceback (most recent call last):
  File "..../llama-agents/zPX/examples/TESTclient.py", line 5, in <module>
    task_result = client.get_task_result(task_id)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pxu/CODE/GenAI4Code/llama-agents/.venv/lib/python3.12/site-packages/llama_agents/client/sync_client.py", line 135, in get_task_result
    return TaskResult(**task.state["result"])
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: llama_agents.types.TaskResult() argument after ** must be a mapping, not NoneType

Original Code
`
from llama_agents import LlamaAgentsClient

client = LlamaAgentsClient("http://0.0.0.0:8001")
task_id = client.create_task("What is the secret fact?")
task_result = client.get_task_result(task_id)
print(task_result.result)
`

Simple fix: to add execution delay:
`
import time

task_id = client.create_task("What is the secret fact?")
time.sleep(10)
task_result = client.get_task_result(task_id)
`

nerdai:
That sounds good. Do you want to open a PR for it?

styk-tv:
please do not introduce static delays, that is absolutely silly. ideally this would be called async from client. 

However, in documentation it says `task_result` should come back None if task_id is not ready. this is a bug @logan-markewich as correctly stated by @peteryxu since TaskResult is not accepting (**None) as parameter.

![image](https://github.com/run-llama/llama-agents/assets/838921/95ad3feb-37d5-4b30-9af4-61668f02e7a1)

logan-markewich:
Thats what the try/except is for -- it raises an error saying that the result isn't ready. The code sample should probably either use a sleep or a try-except.

(Originally I had it returning `None`, but thought raising an error is a bit more informative)

logan-markewich:
@peteryxu was suggesting adding a delay for people who just copy-pasted the code example, not in the actual source code

styk-tv:
@logan-markewich in `async_client.py`

![image](https://github.com/run-llama/llama-agents/assets/838921/e01e754d-002b-4226-91b9-ee60f00b22a2)

[...]

Issue #92, opened 2024-07-03
Installation Instruction missing one pip library


peteryxu:
started from scratch, cloned this repo, created an venv,  and followed instruction doc below:
pip install llama-agents
#if you don't already have llama-index installed, to follow these examples, you'll also need
pip install llama-index-agent-openai

When running agentic_local_single.py, still get error below:
ImportError: `llama-index-embeddings-openai` package not found, please run `pip install llama-index-embeddings-openai`

The Installation section of Readme.md should include the following: 
pip install llama-index-embeddings-openai

logan-markewich:
fair enough, llama-index is defaulting to openai embeddings

Related to #87 this should be more lazily initialized I think

peteryxu:
closed as fixed in PR#106

PR #106, opened 2024-07-05
[docs] Two fixes to provide better out-of-box experience to get started with llama agents

peteryxu:
This PR addresses two issues detailed in links below.  Fixes are both docuement-related in the README.md. The purpose is to allow whoever clones the repo to be able to run samples out of box, without encountering package missing error in #92, or the null async response in #97. 
 
#92 
https://github.com/run-llama/llama-agents/issues/92
#97 
https://github.com/run-llama/llama-agents/issues/97

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ pip install llama-agents
 If you don't already have llama-index installed, to follow these examples, you'll also need
 
 ```bash
-pip install llama-index-agent-openai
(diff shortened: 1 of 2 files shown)


logan-markewich:
Awesome, thanks @peteryxu !

logan-markewich (approved):

PR #111, opened 2024-07-07
Some unit tests for RabbitMQMessageQueue

nerdai:
- Add a few unit tests for RabbitMQMessageQueue (wish there was some test client for RabbitMQ like fastapi provides)

### DIFF
--- a/.github/workflows/unit_test.yml
+++ b/.github/workflows/unit_test.yml
@@ -31,7 +31,7 @@ jobs:
           version: ${{ env.POETRY_VERSION }}
       - name: Install deps
         shell: bash
-        run: poetry install --with dev -E "rabbitmq"
+        run: poetry install --with dev
       - name: Run testing
         env:
           CI: true
--- a/.github/workflows/unit_test.yml
+++ b/.github/workflows/unit_test.yml
@@ -31,7 +31,7 @@ jobs:
           version: ${{ env.POETRY_VERSION }}
       - name: Install deps
         shell: bash
-        run: poetry install --with dev
+        run: poetry install --with dev -E "rabbitmq"
       - name: Run testing
         env:
           CI: true
--- a/tests/message_queues/test_rabbitmq.py
+++ b/tests/message_queues/test_rabbitmq.py
@@ -1,0 +1,21 @@
+from llama_agents.message_queues.rabbitmq import RabbitMQMessageQueue
+
+try:
+    import aio_pika
+except ImportError:
+    aio_pika = None
+
+
+def test_init() -> None:
+    # arrange/act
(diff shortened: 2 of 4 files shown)


.github/workflows/unit_test.yml
```diff
@@ -31,7 +31,7 @@ jobs:
           version: ${{ env.POETRY_VERSION }}
       - name: Install deps
         shell: bash
-        run: poetry install --with dev
+        run: poetry install --with dev -E "rabbitmq"
```
nerdai: I had to add this since tests were failing without `aio_pika` as I "try" to import it in the unit test file. But am a bit confused to be honest cause I wrapped it in try/except block following a similar pattern to what we've done in llama-index. 🤔

tests/message_queues/test_rabbitmq.py
```diff
@@ -0,0 +1,93 @@
+import json
+import pytest
+from llama_agents import QueueMessage
+from llama_agents.message_queues.rabbitmq import RabbitMQMessageQueue
+from unittest.mock import patch, MagicMock, AsyncMock
+
+try:
+    import aio_pika
```
nerdai, line 8: even tho this was in try/except block, the runner was still throwing the exception and halting execution

logan-markewich: Do you think it failed because RabbitMQMessageQueue also imports pika?

.github/workflows/unit_test.yml
logan-markewich: Is aio_pika in the dev dependencies? Should we add it there so unit tests run?

logan-markewich (approved):
LGTM! Just a few small comments

tests/message_queues/test_rabbitmq.py
nerdai: Oh perhaps. This was the error log, so it was just pointing at `aio_pika` but maybe theres another import thats messing it up?

```

==================================== ERRORS ====================================
____________ ERROR collecting tests/message_queues/test_rabbitmq.py ____________
tests/message_queues/test_rabbitmq.py:8: in <module>
    import aio_pika
E   ModuleNotFoundError: No module named 'aio_pika'

During handling of the above exception, another exception occurred:
tests/message_queues/test_rabbitmq.py:12: in <module>
    Exchange = Exchange
E   NameError: name 'Exchange' is not defined
=============================== warnings summary ===============================
```

.github/workflows/unit_test.yml
nerdai: Sure, we can do that. So this dep will exist in two places:

- dev
- extras ("rabbitmq")

Issue #101, opened 2024-07-04
[BUG] Error in executing `examples/pipeline_human_local_single.py`


nerdai:
When trying to run the script `examples/pipeline_human_local_single.py`, I get this error:

```sh
  File "/Users/nerdai/Projects/llama-agents/llama_agents/orchestrators/pipeline.py", line 183, in get_next_messages
    queue_message = get_service_component_message(
  File "/Users/nerdai/Projects/llama-agents/llama_agents/orchestrators/pipeline.py", line 34, in get_service_component_message
    if module.module_type == ModuleType.AGENT:
AttributeError: 'RouterComponent' object has no attribute 'module_type'
```

peteryxu:
ServiceComponent has the attribute of 'module_type', while RouterComponent does not have since it is NOT a  ServiceComponent. 

In the current examples/pipeline_human_local_single.py file, when creating QueryPipeline,  an RouterComponent was passed in, vs ServiceComponent.  

I did some testing, just commented out RouterCompoent, and pass in the two service components directly, and it seems to work in my local env and setup.


####################################
pipeline = QueryPipeline(
    chain=[
        agent_component,
        human_component
        # RouterComponent(
        #     selector=PydanticSingleSelector.from_defaults(llm=OpenAI()),
        #     choices=[agent_service.description, human_service.description],
        #     components=[agent_component, human_component],
        # )
    ]
)

peteryxu:
Tested code above, and it worked as seen below. Created a PR #112 to merge. 

<img width="1026" alt="image" src="https://github.com/run-llama/llama-agents/assets/5605055/175767f5-c0f2-49ff-9546-97ac2aebce5c">

PR #112, opened 2024-07-07
PR for issue #101

peteryxu:
Addressing issue #101

### DIFF
--- a/examples/pipeline_human_local_single.py
+++ b/examples/pipeline_human_local_single.py
@@ -43,11 +43,13 @@ human_component = ServiceComponent.from_service_definition(human_service)
 
 pipeline = QueryPipeline(
     chain=[
-        RouterComponent(
-            selector=PydanticSingleSelector.from_defaults(llm=OpenAI()),
-            choices=[agent_service.description, human_service.description],
-            components=[agent_component, human_component],
-        )
+        agent_component,
+        human_component
+        # RouterComponent(
+        #     selector=PydanticSingleSelector.from_defaults(llm=OpenAI()),
+        #     choices=[agent_service.description, human_service.description],
+        #     components=[agent_component, human_component],
+        # )
     ]
 )
 
--- a/examples/pipeline_human_local_single.py
+++ b/examples/pipeline_human_local_single.py
@@ -10,9 +10,8 @@ from llama_agents import (
 
 from llama_index.core.agent import FunctionCallingAgentWorker
 from llama_index.core.tools import FunctionTool
-from llama_index.core.query_pipeline import RouterComponent, QueryPipeline
+from llama_index.core.query_pipeline import QueryPipeline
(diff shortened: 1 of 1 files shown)


examples/pipeline_human_local_single.py
```diff
@@ -43,11 +42,13 @@ def get_the_secret_fact() -> str:
...
-            choices=[agent_service.description, human_service.description],
-            components=[agent_component, human_component],
-        )
+        agent_component,
+        human_component
+        # RouterComponent(
+        #     selector=PydanticSingleSelector.from_defaults(llm=OpenAI()),
+        #     choices=[agent_service.description, human_service.description],
+        #     components=[agent_component, human_component],
+        # )
```
nerdai, line 51: Thanks @peteryxu. Sadly tho, I don't think this fixes the bug. This replaces the desired `RouterComponent` with a deterministic chain of agent_component > human_component. The desired behaviour is instead just one of these components being chosen to handle a task.

I think we might be missing a RouterServiceComponent that is like RouterComponent but takes ServiceComponents as input and is itself a ServiceComponent.

(cc: @logan-markewich)

nerdai (commented):
Thanks @peteryxu for the attempt to fix the bug. I think this doesn't quite fix the bug, but replaces it with another (which is valid) pipeline but behaves differently from what we intend still.

examples/pipeline_human_local_single.py
logan-markewich: I could have sworn this worked. Perhaps I forgot to commit a change or file. Will try and take a look

peteryxu: > Thanks @peteryxu. Sadly tho, I don't think this fixes the bug. This replaces the desired `RouterComponent` with a deterministic chain of agent_component > human_component. The desired behaviour is instead just one of these components being chosen to handle a task.
> 
> I think we might be missing a RouterServiceComponent that is like RouterComponent but takes ServiceComponents as input and is itself a ServiceComponent.
> 
> (cc: @logan-markewich)

yeah, QueryPipeline expects ServiceComponent in the chain.. while RouterComponent is NOT a service component

peteryxu:
> Thanks @peteryxu for the attempt to fix the bug. I think this doesn't quite fix the bug, but replaces it with another (which is valid) pipeline but behaves differently from what we intend still.

that makes sense.  btw,   pipeline_local_single.py is NOT using RouterComponent.. 
image

examples/pipeline_human_local_single.py
logan-markewich: Rip, I think a change jerry made has made this more complicated than it probably needs to be. Trying to smooth this out locally

logan-markewich:
Going to close this in favour of https://github.com/run-llama/llama-agents/pull/118

PR #122, opened 2024-07-09
Improve summarization logic

ricardo-larosa:
Improve summarization logic, criterions to be defined.

### DIFF
--- a/llama_agents/orchestrators/agent.py
+++ b/llama_agents/orchestrators/agent.py
@@ -1,5 +1,4 @@
-from typing import Any, Dict, List, Tuple
-
+from typing import Any, Dict, List, Optional, Tuple
 from llama_index.core.llms import LLM
 from llama_index.core.memory import ChatMemoryBuffer
 from llama_index.core.tools import BaseTool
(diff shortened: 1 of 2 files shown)


llama_agents/orchestrators/agent.py
```diff
@@ -160,16 +160,17 @@ async def add_result_to_state(
         # summarize the result
         new_history = result.history
         new_history_str = "\n".join([str(x) for x in new_history])
-        # TODO: Better logic for when to summarize?
-        if len(new_history) > 1:
+
+        summary = None
+        if self.should_summarize(new_history):
```
nerdai: I wonder if it might be more simple to just do `summarize()` and then this function always produces a "summary". The current return could just be the string back itself -- in other words, the summary is the string itself for now.

```python
def summarize(self, new_history) -> Optional[str]:
      if len(new_history) > 1:
            new_history_str = "\n".join([str(x) for x in new_history])
            summarize_prompt_str = self.summarize_prompt.format(history=new_history_str)
            return await self.llm.acomplete(summarize_prompt_str)
      return None

optional_summary = self.summarize(new_history)
if optional_summary:
    chat_history.append(..., content=str(optional_summary))
```
```

File: llama_agents/orchestrators/agent.py [hunk 2]
```diff
@@ -160,16 +160,17 @@ async def add_result_to_state(
...
             summarize_prompt_str = self.summarize_prompt.format(history=new_history_str)
             summary = await self.llm.acomplete(summarize_prompt_str)
 
         # get the current chat history, add the summary to it
         chat_dicts = state.get(HISTORY_KEY, [])
         chat_history = [ChatMessage(**x) for x in chat_dicts]
-
-        chat_history.append(ChatMessage(role="assistant", content=str(summary)))
```
nerdai, line 172: I guess this should throw an undefined error if new_history had length 0 or 1?

nerdai (commented):
Thanks @ricardo-larosa -- left a couple of comments. Let me know what you think!

llama_agents/orchestrators/agent.py
ricardo-larosa: I agree, let me know if the way I implemented it makes sense.

llama_agents/orchestrators/agent.py
ricardo-larosa: Let me know if the way I handle this in  `llama_agents/control_plane/server.py` makes sense. I use `ValueError` though or maybe instead of Undefined a custom `EmptyTaskHistoryError` ?

logan-markewich:
With the recent refactor and rename to llama-deploy, this pr is no longer needed.

But really appreciate you taking the time to contribute, and I hope you find the refactor useful!

PR #148, opened 2024-07-21
Add KafkaMessageQueue

### LINKED ISSUE #144: Add Apache Kafka as MessageQueue

nerdai:
This PR adds the `KafkaMessageQueue` which is the integration for Apache Kafka.

- It uses a combination of `kafka-python-ng` (the up-to-date fork of `kafka-python`) as well as `aiokafka` from the same folks who produced `aiohttp`.
- Adds an examples in the `examples/kafka` folder that contain:
     - Single local launch using `KafkaMessageQueue`
     - Without Docker launch using `KafkaMessageQueue`
     - With Docker launch using `KafkaMessageQueue`

closes #144


llama_agents/message_queues/apache_kafka.py
```diff
@@ -0,0 +1,242 @@
...
+        ```
+    """
+
+    url: str = DEFAULT_URL
+
+    def __init__(
+        self,
+        url: str = DEFAULT_URL,
+    ) -> None:
+        super().__init__(url=url)
```
logan-markewich, line 54: I forget, is base message queue pydantic? Do we even need the init here?

llama_agents/message_queues/apache_kafka.py
```diff
@@ -0,0 +1,242 @@
...
+        num_partitions: Optional[int] = None,
+        replication_factor: Optional[int] = None,
+        **kwargs: Dict[str, Any],
+    ) -> None:
+        """Create a new topic.
+
+        Use kafka-python-ng instead of aio-kafka as latter has issues with
+        resolving api_version with broker.
+
+        TODO: convert to aiokafka once this it is resolved there.
```
logan-markewich, line 86: Is there a link to an open issue we can track?

llama_agents/message_queues/apache_kafka.py
```diff
@@ -0,0 +1,242 @@
...
+        """
+        from kafka.admin import KafkaAdminClient, NewTopic
+        from kafka.errors import TopicAlreadyExistsError
```
logan-markewich: Do we need to do a nicer try/except to give the user a better error msg?

llama_agents/message_queues/apache_kafka.py
```diff
@@ -0,0 +1,242 @@
...
+        admin_client = KafkaAdminClient(bootstrap_servers=self.url)
+        active_topics = admin_client.list_topics()
+        topics_to_delete = [el for el in message_types if el in active_topics]
+        admin_client.delete_consumer_groups(DEFAULT_GROUP_ID)
+        if topics_to_delete:
+            admin_client.delete_topics(topics_to_delete)
+
+    async def deregister_consumer(self, consumer: BaseMessageQueueConsumer) -> Any:
+        """Deregister a consumer."""
+        pass
```
logan-markewich, line 140: Should deregister do something, like delete the topic? (Note, not sure how Kafka works at all lol)

logan-markewich, line 158: Should deregister be deleting the topic? Or nah?

examples/kafka/README.md
```diff
@@ -0,0 +1,175 @@
...
+
+```sh
+# using pip install
+pip install llama-agents[kafka]
+
+# using poetry
+poetry add llama-agents -E "kafka"
+```
+
+An virtual environment with this installation of `llama-agents` is what will
```
logan-markewich: *A virtual ..

examples/kafka/pig-latin-translation/pig_latin_translation/additional_services/task_result.py
```diff
@@ -0,0 +1,77 @@
...
+    BaseMessageQueueConsumer,
+    StartConsumingCallable,
+)
+from llama_agents.message_consumers.remote import RemoteMessageConsumer
+from logging import getLogger
+
+logger = getLogger(__name__)
+
+
+class TaskResultService:
```
logan-markewich, line 20: Should we detail this class in the readme? (Also, should it extend the base service class?)

examples/kafka/README.md
logan-markewich, line 1: Future note: Somehow it feels like our examples should be more simple. Feels a little crazy to write this much code for a toy example. I wonder how much of this is worth baking into other abstractions in the library

logan-markewich (approved):
Awesome! Another queue! 💪

llama_agents/message_queues/apache_kafka.py
nerdai: Nah, not in base -- we don't have any fields attached there. Though maybe we should have `url`.

llama_agents/message_queues/apache_kafka.py
nerdai: Good call! There is now: https://github.com/run-llama/llama-agents/issues/149

😊

llama_agents/message_queues/apache_kafka.py
nerdai: Yeah, I think we could check if we can delete the topic perhaps (i.e., if no consumers are subscribed to it). Though this would require the `AdminKafkaClient` again, so I'm thinking we hold off on until the asyncio admin client works.

(I've added it to the issue that tracks the asyncio admin client resolution)

examples/kafka/README.md
nerdai: thanks!

examples/kafka/README.md
nerdai: I agree! I think for simple local launch/single examples its easy.

But when looking to deploy these with some orchestration via docker-compose or kubernetes then I am not sure if it can be as easy as a single script.

It would be really nice to simplify the process of building a multi-agent app where we do "intend" to deploy the microservices on their own.

The pattern I've adopted here is:

- build a multi-agent app Python package
- that defines all of your microservices and launches them in their associated script
- this way we can have a single docker image and just change the command for that image to launch the service that you want to launch

examples/kafka/pig-latin-translation/pig_latin_translation/additional_services/task_result.py
nerdai: So this is equivalent to our `HumanService` in our local launcher examples (I just renamed it and gave its own python module).

I'll add details to the README for sure! 

Though, I'm not sure if it should extend the base service since:

- it's not a `publisher` (and `BaseService` has the `MessageQueuePublisherMixin`)
- and it doesn't need to register to the control plane (since by default finalized tasks get sent to "human" consumer already)

llama_agents/message_queues/apache_kafka.py
nerdai: Are you referring to try/except on the import statements as we do in llama-index?

examples/kafka/pig-latin-translation/pig_latin_translation/additional_services/task_result.py
logan-markewich: That's fair! Thanks for clarifying!

llama_agents/message_queues/apache_kafka.py
nerdai: 👍🏼

examples/kafka/README.md
nerdai: Added an issue to track this for us: https://github.com/run-llama/llama-agents/issues/150 😅

llama_agents/message_queues/apache_kafka.py
nerdai: https://github.com/run-llama/llama-agents/issues/151

Added this as an issue in our board -- with the label of "good first issue".

nerdai:
Alright, thanks for the review @logan-markewich! I'm pretty sure I got everything covered from your comments thru direct changes or adding some new issues in our board!

Going to merge this now. 🙏🚀

PR #163, opened 2024-07-29
docs: updated readme's notebook flow example

### LINKED ISSUE #162: Doesn't work in notebooks
The docs says:
Next, when working in a **notebook** or for faster iteration, we can launch our llama-agents system in a single-run setting, where one message is propagated through the network and returned.
```
from llama_agents import LocalLauncher


# launch it
launcher = LocalLauncher(
    [agent_server_1, agent_server_2],
    control_plane,
    message_queue,
)
result = launcher.launch_single("What is the secret fact?")

print(f"Result: {result}")
```
But output is:
RuntimeError: asyncio.run() cannot be called from a running event loop

tomaszbk:
Please let met know if you have any change suggestion for changes to this PR.

Closes #162

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,7 @@ result = launcher.launch_single("What is the secret fact?")
 print(f"Result: {result}")
 ```
 
+<!-- prettier-ignore -->
 > [!NOTE]
 > `launcher.launch_single` creates a new asyncio event loop. Since Jupyter notebooks already have an event loop running, we need to use `nest_asyncio` to allow the creation of new event loops within the existing one.
 
--- a/README.md
+++ b/README.md
@@ -85,6 +85,10 @@ Next, when working in a notebook or for faster iteration, we can launch our `lla
 
 ```python
 from llama_agents import LocalLauncher
+import nest_asyncio
+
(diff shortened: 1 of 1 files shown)


tomaszbk:
removed whitespaces and formatted with prettier

tomaszbk:
Precommit fails and expects this format:
> [!NOTE] > `launcher.launch_single` creates a new asyncio event loop. Since Jupyter notebooks already have an event loop running, we need to use `nest_asyncio` to allow the creation of new event loops within the existing one

While the format I provide is this:
> [!NOTE]
> `launcher.launch_single` creates a new asyncio event loop. Since Jupyter notebooks already have an event loop running, we need to use `nest_asyncio` to allow the creation of new event loops within the existing one.

nerdai:
Looks like `prettier` not so happy with that -- https://github.com/prettier/prettier/issues/15479.

I made the change as suggested in there to ignore for now:

```md
<!-- prettier-ignore -->
...
```

nerdai (approved):
lgtm! thanks @tomaszbk 🙏🏼

PR #164, opened 2024-08-02
Use other sources for agent orchestrator summary

micahparker:
ATM, the summary variable can go unset and throw an exception causing an infinite loop of deque pop errors. Im not sure these are the right sources for the variable, obv, but it prevents the exception.


logan-markewich (approved):
This makes sense to me. Probably room to modify that "default summary" but this works for now

PR #166, opened 2024-08-03
Add RouterOrchestrator

nerdai:
Similar to `RouterComponent`, this PR adds `RouterOrchestrator` which selects an appropriate orchestrator from a set of choices for a given task (i.e. input).

### DIFF
--- a/llama_agents/orchestrators/router.py
+++ b/llama_agents/orchestrators/router.py
@@ -13,7 +13,58 @@ logger = logging.getLogger(__name__)
 
 
 class RouterOrchestrator(BaseOrchestrator):
-    """Orchestrator that routes between a list of pipeline orchestrators."""
+    """Orchestrator that routes between a list of orchestrators.
+
+    Given an incoming task, first select the most relevant orchestrator to the
+    task, and then use that orchestrator to process it.
+
+    Attributes:
+        orchestrators (List[BaseOrchestrator]): The orchestrators to choose from. (must correspond to choices)
+        choices (List[str]): The descriptions of the orchestrators (must correspond to components)
+        selector (BaseSelector): The orchestrator selector.
+
+    Examples:
+        ```python
+        from llama_index.core.query_pipeline import QueryPipeline
+        from llama_agents import (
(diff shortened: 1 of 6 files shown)


llama_agents/orchestrators/router.py
```diff
@@ -0,0 +1,103 @@
...
+        selector: BaseSelector,
+    ):
+        self.orchestrators = orchestrators
+        self.choices = choices
+        self.selector = selector
+        self.tasks: Dict[str, int] = {}
+
+    def _select_orchestrator(self, task_def: TaskDefinition) -> BaseOrchestrator:
+        if task_def.task_id not in self.tasks:
+            sel_output = self.selector.select(self.choices, task_def.input)
```
logan-markewich: we should be using async `await self.selector.aselect(..)`

llama_agents/orchestrators/router.py
```diff
@@ -0,0 +1,103 @@
...
+        orchestrator = RouterOrchestrator(
+            selector=PydanticSingleSelector.from_defaults(llm=OpenAI()),
+            orchestrators=[orchestrator_1, orchestrator_2],
+            choices=["description of orchestrator_1", "description of orchestrator_2"],
+        )
+    """
+
+    def __init__(
+        self,
+        orchestrators: List[BaseOrchestrator],
```
logan-markewich: Oh interesting, this isn't routing between services, but routing between orchestrators 👀

nerdai: Yeah. I tried using routing with services, but the issue was I had something like this:

```python
outer_pipeline = QueryPipeline(
    chain = [inner_pipeline, component]
)
orchestrator = PipelineOrchestrator(outer_pipeline)
```

The `inner_pipeline` doesn't get orchestrated properly. It just gets run to completion rather than being properly orchestrated with messages being sent to the message queue.

llama_agents/orchestrators/router.py
nerdai: done - thanks!

PR #170, opened 2024-08-08
WorkflowService

logan-markewich:
This PR introduces a workflow service, for deploying workflows as services

It makes a few key assumptions
- workflows are separated from their state
- the state is passed in when running the service
- the state of the workflow (i.e. the context) is currently **pickled** 
    - we could avoid pickles if we ensure that the context can only hold basic serializable objects
    - currently I try to add some minor security by using salted hashes on the pickle string. Still not great


llama_agents/services/workflow.py
```diff
@@ -0,0 +1,340 @@
...
+logger = getLogger(__name__)
+hash_secret = str(os.environ.get("LLAMA_AGENTS_HASH_SECRET", "default"))
+
+
+class WorkflowState(BaseModel):
+    """Holds the state of the workflow.
+
+    Used to validate and pass message payloads.
+
+    TODO: Should this be the general payload for all messages?
```
nerdai, line 41: I guess this would make sense if we're moving to a world where all agents/humans services are executing a step in a Workflow? Or am I misinterpreting something here?

llama_agents/services/workflow.py
```diff
@@ -0,0 +1,340 @@
...
+        context_bytes = pickle.dumps(context)
+        context_str = base64.b64encode(context_bytes).decode("ascii")
+        context_hash = hash(context_str + hash_secret)
+
+        return WorkflowState(hash=context_hash, state=context_str, run_kwargs={})
+
+    async def processing_loop(self) -> None:
+        """The processing loop for the service.
+
+        TODO: How do we handle any errors that occur during processing?
```
nerdai, line 221: Good question...maybe in the future its something like message stays in the MessageQueue until a consumer (WorkflowService) successfully processes it. I think we get that with some of our message queue integrations.

Though maybe we want some error/retry logic in a single processing attempt?

nerdai (commented):
The `WorkflowService` looks good to me. I'll test it out on my machine to see why we're running into issues with `instrumentation` here.

llama_agents/services/workflow.py
```diff
@@ -0,0 +1,340 @@
...
+            if not self.running:
+                await asyncio.sleep(self.step_interval)
+                continue
+
+            async with self.lock:
+                current_calls = [(t, c) for t, c in self._outstanding_calls.items()]
+
+            for task_id, current_call in current_calls:
+                # "fork" the workflow, clear its state
+                self.workflow._queues = {}
```
erik-megarad, line 234: I think we should create a new object instead of modifying properties on the existing workflow object.

llama_agents/services/workflow.py
```diff
@@ -0,0 +1,340 @@
...
+                            history=[],
+                            result=str(result),
+                            data=updated_state.dict(),
+                        ).model_dump(),
+                    )
+                )
+
+                # clean up
+                async with self.lock:
+                    del self._outstanding_calls[task_id]
```
erik-megarad: If for whatever reason the `task_id` is not in the dict this will raise a `KeyError`. The preferred approach would be to do:

```suggestion
                    self._outstanding_calls.pop("task_id", None)
```

llama_agents/services/workflow.py
```diff
@@ -0,0 +1,340 @@
...
+        """Get the consumer for the message queue.
+
+        Args:
+            remote (bool):
+                Whether the consumer is remote. Defaults to False.
+                If True, the consumer will be a RemoteMessageConsumer that uses the `process_message` endpoint.
+        """
+        if remote:
+            url = (
+                f"http://{self.host}:{self.port}{self._app.url_path_for('process_message')}"
```
erik-megarad, line 294: Should we be more aware of the protocol? I imagine wanting `https` would be common.

llama_agents/services/workflow.py
logan-markewich: I agree, we just don't have a good method for doing that just yet

llama_agents/services/workflow.py
logan-markewich: That makes sense. We might also want a FAILURE message type too

logan-markewich: Then the caller can decide retry logic?

llama_agents/services/workflow.py
logan-markewich: I think here i meant generalizing the way we pass data to services 

Tbh the task definition almost does this

llama_agents/services/workflow.py
logan-markewich: I agree, it should probably be an attribute on the service class

llama_agents/services/workflow.py
nerdai: Yeah I think FAILURE message type makes sense to have. I like passing the decision to the caller once they receive this failure message. 👍🏼

llama_agents/services/workflow.py
nerdai: yeah, we almost always i think pass `data` as `TaskDefinition.model_dump()`

PR #191, opened 2024-08-17
Introduce session concept to the control plane

logan-markewich:
This PR turned out to be a decent refactor.

Basically, the idea is to link tasks together in a session, and have the session manage the state rather than the task itself.

The usecase here is the services can be stateful (like a workflow implementation of a react agent), and you need to maintain the state between runs.

So in order to let the service scale, the state is managed externally from the service in the control plane. 

The UX here is that the user creates a session, and then can have several back/forths with the service, with the state for those interactions being managed by the session.

A (mostly?) full changelist
- Removed the `state` attribute for `TaskDefinition`
- Introduced `SessionDefinition` to hold the state and task ids for that session 
- Introduced `NewTask` to handle passing state + task to services (since the state is not directly attached to the task now)
- Refactored the control plane interface to account for sessions
- Removed the "tools" argument from the orchestrator base class
- Deleted the old orchestrator classes (We can add these back in another form later if we think they are still helpful)
- Added and fixed tests


llama_agents/types.py
```diff
@@ -86,21 +86,53 @@ class TaskDefinition(BaseModel):
...
+            return None
+
+        return self.task_ids[-1]
+
+
+class NewTask(BaseModel):
+    """The payload for a new task message."""
+
+    task: TaskDefinition
+    state: Dict[str, Any] = Field(default_factory=dict)
```
jerryjliu, line 133: lack context, but is NewTask needed just to pass state around? If TaskDefinition stores session_id, and there's a general way to look up the session state by session_id and modify it, might be better? less of a risk the state in NewTask and state in the Session are out of sync

llama_agents/control_plane/server.py
```diff
@@ -167,31 +165,16 @@ def publisher_id(self) -> str:
...
-    async def get_total_services(self) -> int:
-        all_services = await self.state_store.aget_all(
-            collection=self.services_store_key
-        )
-        return len(all_services)
-
     async def process_message(self, message: QueueMessage) -> None:
         action = message.action
 
         if action == ActionTypes.NEW_TASK and message.data is not None:
```
jerryjliu, line 171: again lack context but i wonder if we should add NEW_SESSION as an explicit action. for instance user on the client side might want to explicitly create a new session

logan-markewich: There's a new session endpoint on the control plane, which will be the main point for creating a session from the client 👍

llama_agents/types.py
logan-markewich: Fair point. How it works right now is the control plane has the details for how state is stored, and both the orchestrator and the service modify the state, before the control plane saves it again (State is stored using one of our kvstore abstractions)

I could give the details to look up the state from the services, but it also feels kind of nice that only the control plane can touch the state :thinking:

logan-markewich: Tbh thinking about this now, there might need to be a lock somehow while the state is checked out. Or something like that (lower priority imo)

llama_agents/control_plane/base.py
```diff
@@ -105,33 +133,73 @@ async def handle_service_completion(
...
+        Get all sessions.
 
         Returns:
-            dict: State of the task
+            dict: All sessions, mapped from session ID to session definition.
         """
         ...
 
     @abstractmethod
-    async def get_all_tasks(self) -> dict:
```
nerdai, line 129: I guess there really wouldn't be any sensible use for getting all tasks across all sessions?

llama_agents/control_plane/base.py
```diff
@@ -105,33 +133,73 @@ async def handle_service_completion(
...
+        ...
+
+    @abstractmethod
+    async def get_task(self, task_id: str) -> TaskDefinition:
+        """
+        Get the specified task.
+
+        Args:
+            task_id (str): Unique identifier of the task.
+            session_id (str): Unique identifier of the session.
```
nerdai: are we missing the `session_id` in the function signature?

llama_agents/control_plane/server.py
```diff
@@ -275,65 +242,117 @@ async def get_all_services(self) -> Dict[str, ServiceDefinition]:
...
+
+    async def get_session_tasks(self, session_id: str) -> List[TaskDefinition]:
+        session = await self.get_session(session_id)
+        task_defs = []
+        for task_id in session.task_ids:
+            task_defs.append(await self.get_task(task_id))
+        return task_defs
+
+    async def get_current_task(self, session_id: str) -> Optional[TaskDefinition]:
+        session = await self.get_session(session_id)
```
nerdai, line 288: Would there be a time where multiple tasks can be running at the same time? In that case, `get_current_task` would/should return all running tasks?

llama_agents/message_consumers/callable.py
```diff
@@ -14,6 +14,7 @@ class CallableMessageConsumer(BaseMessageQueueConsumer):
     handler: Callable
 
     async def _process_message(self, message: QueueMessage, **kwargs: Any) -> None:
+        print("In callable handler")
```
nerdai: I think maybe this is leftover from debugging/testing process?

llama_agents/orchestrators/base.py
```diff
@@ -12,15 +10,13 @@ class BaseOrchestrator(ABC):
 
     The general idea for an orchestrator is to manage the flow of messages between services.
 
-    Given some state, task, and list of tools, figure out the next messages to publish. Then, once
```
nerdai, line 15: Before, we had wrapped tools as a `Service` and this is we used function calling for a way to route to the services through this, right?

llama_agents/types.py
nerdai: yeah agreed, probably need to lock to ensure most up to date version is being read, and no data races with writing. 👍

nerdai (approved):
I like this concept of teasing out state from tasks and using sessions! The only thing missing for me right now is that we've removed a few of our Orchestrators that defined the execution of a Task (now "Session"?) thru the multi-agent system. How are we envisioning orchestrators do that now... or are we removing that responsbility to something else?

llama_agents/control_plane/base.py
logan-markewich: yea I couldn't quite think of a reason to expose that

llama_agents/control_plane/base.py
logan-markewich: hmm, let me double check, this might be leftover copy-pasta

llama_agents/control_plane/server.py
logan-markewich: yea its a good point. I might need to introduce some kind of status state to the tasks 😬

llama_agents/message_consumers/callable.py
logan-markewich: yes, thank you!

llama_agents/control_plane/base.py
logan-markewich: This is indeed leftover copy-pasta -- the tasks are stored with the task_id as the key, but you need the session id already to get the task ids

We could make the key `session_id + task_id`, but tbh I don't see why we need to complicate that further

llama_agents/orchestrators/base.py
logan-markewich: Right. But now, its up to each service to route/call between services, and the top-level SimpleOrchestrator handles generic retries, failures, etc.

llama_agents/control_plane/server.py
logan-markewich: Job state is more in depth than i thought. Moving this to its own ticket for now

PR #199, opened 2024-08-22
initial service deployment

logan-markewich:
This PR does the following
- removes the old launchers
- introduces a new launcher
    - async only
    - debatable if this should have sync entry points 
- introduces config objects for each type of services
    - message queue
    - control plane
    - workflow service
    - orchestrator
    - debatable if this is ok to do
- adds some missing methods to the control plane

----

Tested with the following code:

script1.py
```python
from llama_agents import deploy_core, ControlPlaneConfig, SimpleMessageQueueConfig

async def main():
    await deploy_core(
        ControlPlaneConfig(),
        SimpleMessageQueueConfig(),
    )

if __name__ == "__main__":
    import asyncio

    loop = asyncio.get_event_loop()
    loop.run_until_complete(main())
```

script2.py
```python
from llama_index.core.workflow import Workflow, StartEvent, StopEvent, step

class MyWorkflow(Workflow):

    @step()
    async def run_step(self, ev: StartEvent) -> StopEvent:
        arg1 = ev.get("arg1")
        if not arg1:

from llama_agents import (
    deploy_workflow, 
    ControlPlaneConfig, 
    SimpleMessageQueueConfig, 
    WorkflowServiceConfig,
)

async def main():
    await deploy_workflow(
        MyWorkflow(),
        WorkflowServiceConfig(host="127.0.0.1", port=8002, service_name="my_workflow"),
        ControlPlaneConfig(),
        SimpleMessageQueueConfig(),
    )

if __name__ == "__main__":
    import asyncio
    
    loop = asyncio.get_event_loop()
    loop.run_until_complete(main())
```

(avoiding asyncio.run() makes `ctrl+c` exit the script a lot more elegantly 😅)


llama_agents/message_queues/simple.py
```diff
@@ -29,6 +29,13 @@
 logger = getLogger(__name__)
 
 
+class SimpleMessageQueueConfig(BaseModel):
```
nerdai: Do we want to use `BaseSettings` ([source](https://docs.pydantic.dev/latest/concepts/pydantic_settings/#usage)) instead?

I did something [similar](https://github.com/run-llama/llama_index/blob/ef9a21c73d35f01332cdbfece5a31f4125e2a939/llama-index-networks/llama_index/networks/contributor/retriever/service.py#L11) in networks

llama_agents/launchers/client.py
```diff
@@ -0,0 +1,161 @@
...
+            for task in all_tasks:
+                if task.done() and task.exception():  # type: ignore
+                    raise task.exception()  # type: ignore
+
+    async def deploy_workflow(
+        self,
+        workflow: Workflow,
+        workflow_config: WorkflowServiceConfig,
+        control_plane_config: ControlPlaneConfig,
+        message_queue_config: BaseModel,
```
nerdai: I like this way better than before, because we were passing the control plane and services objects where we could've just passed the service params like we do with config.

nerdai (approved):
Took a quick a look. UX looks improved from before to me.

llama_agents/message_queues/simple.py
logan-markewich: Added!

logan-markewich:
Added BaseSettings, moved the deploy functions to be free-functions. Going to merge as-is :)

PR #202, opened 2024-08-26
Make client more useful/up to date

logan-markewich:
This PR
- re-implementes the clients (async and sync) to match the latest control plane endpints
- introduces a session-based client to make API calls and session separation clearer

Unknown:
- I'd love to add unit tests for the client, but I'm not sure how to do this without mocking out the entire control plane API 😢 

### `launch_core.py`

```python
from llama_agents import deploy_core, ControlPlaneConfig, SimpleMessageQueueConfig

async def main():
    await deploy_core(
        ControlPlaneConfig(),
        SimpleMessageQueueConfig(),
    )

if __name__ == "__main__":
    import asyncio

    loop = asyncio.get_event_loop()
    loop.run_until_complete(main())
```

### `launch_workflow.py`

```python
from llama_index.core.workflow import Workflow, StartEvent, StopEvent, step

class MyWorkflow(Workflow):

    @step()
    async def run_step(self, ev: StartEvent) -> StopEvent:
        arg1 = ev.get("arg1")
        if not arg1:

from llama_agents import deploy_workflow, ControlPlaneConfig, SimpleMessageQueueConfig, WorkflowServiceConfig

async def main():
    await deploy_workflow(
        MyWorkflow(),
        WorkflowServiceConfig(host="127.0.0.1", port=8002, service_name="my_workflow"),
        ControlPlaneConfig(),
        SimpleMessageQueueConfig(),
    )

if __name__ == "__main__":
    import asyncio
    
    loop = asyncio.get_event_loop()
    loop.run_until_complete(main())
```

### `test_client.py`

```python
from llama_agents import LlamaAgentsClient

# points to control plane URL
client = LlamaAgentsClient(ControlPlaneConfig())

session = client.create_session()
result = session.run("my_workflow", arg1="hello_world")
print(result)
> 'hello_world_result'
```


nerdai:
Quick thought came to mind: is it necessary to separate core stuff: `ControlPlaneConfig` and `MessageQueueConfig`. Or, would just have one `CoreConfig`?

llama_agents/client/async_client.py
```diff
@@ -1,85 +1,186 @@
...
-    async def create_task(self, task_def: Union[str, TaskDefinition]) -> str:
-        """Create a new task with the control plane.
+    async def run(self, service_name: str, **run_kwargs: Dict[str, Any]) -> str:
+        """Implements the workflow-based run API for a session."""
+        task_input = json.dumps(run_kwargs)
+        task_def = TaskDefinition(input=task_input, agent_id=service_name)
+        task_id = await self.create_task(task_def)
+
+        # wait for task to complete, up to timeout seconds
+        start_time = asyncio.get_event_loop().time()
```
nerdai: I think we could also use instead here `asyncio.wait_for(coro=..., timeout=...)`. The `coro` can do the polling, similar to what we have for `ServiceAsTool`

https://github.com/run-llama/llama-agents/blob/89e0fd70a580babcd48b0604c0c9c686b7d34d95/llama_agents/tools/service_as_tool.py#L253

llama_agents/client/async_client.py
```diff
@@ -1,85 +1,186 @@
...
+                f"{self.control_plane_url}/sessions/{self.session_id}/tasks/{task_id}/result"
+            )
+            data = response.json()
+            return TaskResult(**data) if data else None
 
-    async def get_services(self) -> Dict[str, ServiceDefinition]:
-        """Get all services registered with the control plane.
+
+class AsyncLlamaAgentsClient:
+    def __init__(self, control_plane_url: str, timeout: float = DEFAULT_TIMEOUT):
```
nerdai: Should this maybe be `ControlPlaneConfig`? Assuming that we plan to store even secrets in such `Config`s?

llama_agents/control_plane/server.py
```diff
@@ -396,10 +414,43 @@ async def get_task(self, task_id: str) -> TaskDefinition:
...
+        if not isinstance(result, TaskResult):
+            if isinstance(result, dict):
+                result = TaskResult(**result)
+            elif isinstance(result, str):
+                result = TaskResult(**json.loads(result))
+            else:
+                raise HTTPException(status_code=500, detail="Unexpected result type")
+
+        # sanity check
+        if result.task_id != task_id:
```
nerdai, line 449: should we insert a log statement here in case it may be useful for future inspections?

nerdai (commented):
Generally looks pretty good! Left a couple of minor-ish comments (I think). And, in terms of unit tests, would `fastapi.testclient` make life a bit easier and not have to mock the entire control plane? I think you can just mock specific requests at a time.

https://fastapi.tiangolo.com/tutorial/testing/

logan-markewich:
@nerdai there could be a core config, but was trying to make things flatter. But if nesting the message queue config into another class feels more natural too, I could consider it 👀

Will look into tests with that testing library!

nerdai:
> @nerdai there could be a core config, but was trying to make things flatter. But if nesting the message queue config into another class feels more natural too, I could consider it 👀
> 
> 
> 
> Will look into tests with that testing library!

Ah fair point. I'm not sure if it is more natural... maybe we should leave as is!

llama_agents/client/async_client.py
logan-markewich: added

llama_agents/control_plane/server.py
logan-markewich: done

llama_agents/client/async_client.py
logan-markewich: done

nerdai (approved):
lgtm!

PR #208, opened 2024-08-29
Add queue config as an endpoint to the control plane

logan-markewich:
Adding the queue config as a GET api on the control plane

This will simplify deployment and imports for the user.


llama_agents/control_plane/server.py
```diff
@@ -122,6 +129,12 @@ def __init__(
             methods=["POST"],
             tags=["Control Plane"],
         )
+        self.app.add_api_route(
+            "/queue_config",
+            self.get_message_queue_config,
+            methods=["GET"],
+            tags=["Control Plane"],
```
nerdai: nit: should this be "Message Queue"? I don't have a strong opinion either way. Not sure if we'll need any other endpoints for the Message Queue in the future?

nerdai (approved):
lgtm!

llama_agents/control_plane/server.py
logan-markewich: Fair enough!

PR #214, opened 2024-09-03
More docs

logan-markewich:
- renamed LlamaAgentsClient to LlamaDeployClient
- updated readme
- copied readme into a module guides page (to be copied into llama-index docs builds)


seldo (approved):
Looks good.

PR #227, opened 2024-09-05
Logan/disable deployments

logan-markewich:


llama_deploy/deploy/deploy.py
```diff
@@ -114,28 +118,35 @@ async def deploy_core(
...
-    message_queue_task = None
-    if isinstance(message_queue_config, SimpleMessageQueueConfig):
+    if (
+        isinstance(message_queue_config, SimpleMessageQueueConfig)
+        and not disable_message_queue
+    ):
         message_queue_task = _deploy_local_message_queue(message_queue_config)
+    else:
+        # create a dummy task to keep the event loop running
+        message_queue_task = asyncio.create_task(asyncio.sleep(0))
```
nerdai, line 131: i was like hang on how did we define this.... sneaky good! lol

pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.poetry]
 name = "llama-deploy"
-version = "0.1.0b10"
+version = "0.1.0b14"
```
nerdai: i think we're at 15 now.

nerdai (approved):
lgtm! only thing is i think we're already at `v0.1.0b14` so we need to bump this up by one

PR #245, opened 2024-09-09
Add Docker machinery for ready-made images

masci:
This PR introduces the machinery needed to build (and possibly publish) basic docker images for the control plane and the message queue. This would greatly simplify the deployment of custom workflows, which would not need to carry around the python code needed to setup llama_deploy. For example, the full-stack example would go from this:
```
├── README.md
├── docker-compose.yml
├── llama_deploy_app
│   ├── control_plane
│   │   ├── deploy.py
│   │   ├── dockerfile
│   │   └── requirements.txt
│   ├── frontend
│   │   ├── assets
│   │   │   └── favicon.ico
│   │   ├── dockerfile
│   │   ├── frontend
│   │   │   ├── __init__.py
│   │   │   ├── frontend.py
│   │   │   ├── state.py
│   │   │   └── style.py
│   │   ├── requirements.txt
│   │   └── rxconfig.py
│   ├── message_queue
│   │   ├── deploy.py
│   │   ├── dockerfile
│   │   └── requirements.txt
│   └── workflows
│       ├── agent_workflow.py
│       ├── data
│       │   └── attention.pdf
│       ├── deploy.py
│       ├── dockerfile
│       ├── rag_workflow.py
│       └── requirements.txt
└── llama_deploy_frontend.png
```
to this
```
├── README.md
├── docker-compose.yml
├── frontend
│   ├── assets
│   │   └── favicon.ico
│   ├── dockerfile
│   ├── frontend
│   │   ├── __init__.py
│   │   ├── frontend.py
│   │   ├── state.py
│   │   └── style.py
│   ├── requirements.txt
│   └── rxconfig.py
├── llama_deploy_frontend.png
└── workflows
    ├── agent_workflow.py
    ├── data
    │   └── attention.pdf
    ├── deploy.py
    ├── dockerfile
    ├── rag_workflow.py
    └── requirements.txt
```
Additionally, we could provide a pre-defined `docker-compose.yml` file containing the boilerplate needed for running llama_deploy so that users can just do in their compose files (if this is interesting, we can find a way to easily make available the boilerplate compose file):
```
  control_plane:
    extends:
      file: ../../docker/docker-compose.yml
      service: message_queue
```


examples/python_fullstack/docker-compose.yml
```diff
@@ -1,6 +1,14 @@
...
 services:
+  message_queue:
+    extends:
+      file: ../../docker/docker-compose.yml
+      service: message_queue
+
+  control_plane:
+    extends:
+      file: ../../docker/docker-compose.yml
+      service: message_queue
```
logan-markewich: typo: should be `control_plane` I think

logan-markewich: I'm a little unfamiliar with this syntax for docker compose -- I'm assuming the user can still override stuff like the ports and hosts?

docker/Dockerfile.base
logan-markewich, line 1: This folder could use a readme 🙏🏻

docker/docker-compose.yml
```diff
@@ -0,0 +1,37 @@
+services:
+  message_queue:
```
nerdai: how should we do this for the other message queues (rabbitmq, redis, kafka, etc.)

nerdai (commented):
I like this a lot! I was considering potentially using [Fragments](https://docs.docker.com/reference/compose-file/fragments/) and Extensions in the message-queue-integrations example, but am not yet too familiar with them.

examples/python_fullstack/docker-compose.yml
masci: Yes, you inherit then you can override/integrate see https://docs.docker.com/compose/multiple-compose-files/extends/

docker/docker-compose.yml
masci: We can make the entrypoint script `message_queue.py` more clever and pass configurations through env vars, so that the compose file is in charge. Then I would probably provide ad-hoc compose files so that users would already have the required service for a certain message queue up and running.

docker/Dockerfile.base
masci: added!

docker/docker-compose.yml
masci: Done, the message queue can be configured passing in the env var `MESSAGE_QUEUE_CONFIG` with one of the values `(kafka, rabbitmq, redis, simple)`

examples/python_fullstack/frontend/requirements.txt
```diff
@@ -0,0 +1,2 @@
+llama-deploy[redis]==0.1.1
```
nerdai: this was from my commit, but can we change this back to ">="

docker/docker-compose-redis.yml
nerdai, line 1: do we want to make docker-compose files for kafka and rabbitmq as well?

docker/README.md
```diff
@@ -0,0 +1,59 @@
...
+multiple platforms feature is currently not supported for docker driver. Please switch to a different driver
+(eg. “docker buildx create --use”)
+```
+
+To work around this problem, one solution is to override the `platform` option and
+limit local builds to the same architecture as your computer's. For example, on an Apple M1 you can limit the builds to ARM only by invoking `bake` like this:
+
+```sh
+docker buildx bake control_plane --set "*.platform=linux/arm64"
+```
```
nerdai, line 49: If using Docker Desktop, which was the case for me, I ended up enabling `containerd image store` (see [here](https://docs.docker.com/build/building/multi-platform/#enable-the-containerd-image-store))

nerdai (approved):
looks great! left a few minor comments

examples/python_fullstack/frontend/requirements.txt
logan-markewich: why does the frontend need llama-deploy[redis] ? I think it would only be `llama-deploy` since it just needs the client

logan-markewich:
This might be a silly question, but I can't deploy the example now?

```
python_fullstack % docker compose up
service "agentic_workflow" depends on undefined service control_plane: invalid compose project
```

Was making sure the README still made sense and ran into this. Did I miss a step?

logan-markewich:
(Also @masci for some reason I can't push to the PR 😅 The readme for the python full stack example is out of date though)

examples/python_fullstack/frontend/requirements.txt
nerdai: Yeah that's a good point. I am pretty certain I ran into an error that led me to install the extra on the frontend as well. Will try to repro now.

nerdai:
> This might be a silly question, but I can't deploy the example now?
> 
> ```
> python_fullstack % docker compose up
> service "agentic_workflow" depends on undefined service control_plane: invalid compose project
> ```
> 
> Was making sure the README still made sense and ran into this. Did I miss a step?

@logan-markewich:

Hmmm... that is very strange. Maybe something before that prevented the control plane from being spun up? Did you run `docker buildx bake` in the `docker` folder?

examples/python_fullstack/frontend/requirements.txt
nerdai: Nope. I was wrong it seems — you don't need this optional with frontend. 

@masci: please kindly remove 🙏🏼

logan-markewich:
I did not run bake 👀 I'll try that (and if that's needed, let's get that into the readme 💪)

masci:
@nerdai @logan-markewich to be clear, before merging this PR I would publish the `llama_deploy` docker image and amend the compose files in this changeset accordingly, so images are pulled from docker hub. The whole point is to have users do less steps so there should be no need to build the images to run the example.

masci:
> (Also @masci for some reason I can't push to the PR 😅 The readme for the python full stack example is out of date though)

No idea how this could be even done, what's the error you get?

masci:
The compose files now pull a non-existent version `0.1.2` from docker hub, we would need a bugfix release after merging this PR.

logan-markewich:
@masci do you want to setup the docker publishing machinery? We can probably add it as a github action?

masci:
> @masci do you want to setup the docker publishing machinery? We can probably add it as a github action?

It's there already, we can merge as it is but we would need a release, or I can pull the `main` tag in the examples and it'll work right on merge.

logan-markewich:
ah I see -- I can make a release

logan-markewich (approved):

PR #161, opened 2024-07-27
AWS SQS MessageQueue Integration

nerdai:
This originally came from #158 but @nerdai made a silly error somehow and merged it into main prematurely. It has since been reverted from the main repo.

This is the PR re-opened.

- Adds SQSMessageQueue
- Uses AWS SNS and SQS

(Cc @MartinRistov)

PR #246, opened 2024-09-09
[Feature] AWSMessageQueue (llama-deploy)

### LINKED ISSUE #251: [Feature Request] Implement `AWSMessageQueue`

nerdai:
Closes #161 which is the PR that aimed to bring `AWSMessageQueue` prior to `v0.1.0` release.

Credit: @MartinRistov
Closes #251


nerdai:
@MartinRistov, I had to re-jig a couple of minor-natured things (i.e with the config and the message queue integration) but I've tested it and everything now works :)

As with your error:

```sh
control_plane-1        |   File "/app/multi_workflows_app/deployment/core.py", line 10, in <module>
control_plane-1        |     from llama_deploy.message_queues.aws import AWSMessageQueueConfig
control_plane-1        | ModuleNotFoundError: No module named 'llama_deploy.message_queues.aws'
Gracefully stopping... (press Ctrl+C again to force)
dependency failed to start: container message-queue-integrations-control_plane-1 exited (1)
```

This is because we need the docker image is not building llama-deploy with your most recent additions. We gotta still figure out the best way to enable outside collaborators to do this, as what I did was I published a pre-release version of our llama-deploy that had our new AWSMessageQueue.

nerdai:
@MartinRistov:
I think there are only a couple of things left to do (unless I am missing anything):

1. Instead of a README as originally planned, I think it may be better at this point to add that information under the class doc string for `AWSMessageQueue`. So, if you feel that docstring needs beefing up, please feel free to add content there.
2. `AWSMessageQueueConfig` reads in the secret creds as regular str types. This is a bit problematic because `control_plane` server has a GET `/queue_config` endpoint that returns the message queue config (json). Currently this shows the secrets in plain text in the payload and thus is insecure. We should re-jig this to be a SecretStr instead. I'm going to work on this now.

nerdai:
Alright re: item 2, I ran into some issues on the FastAPI side with serializing `SecretStr`. So, I am just excluding the secrets from being serialized and only adding them at construction time via the `get_credentials()` method of the config.

llama_deploy/message_queues/aws.py
```diff
@@ -0,0 +1,381 @@
...
+        super().__init__(**kwargs)
+        self._aio_session = None
+
+        # AWS retry configuration with exponential backoff
+        self._retry_config = Config(retries={"max_attempts": 5, "mode": "adaptive"})
+
+        if not self.aws_region:
+            raise ValueError("AWS region must be provided.")
+
+        if not self.aws_access_key_id and not self.aws_secret_access_key:
```
nerdai, line 107: @MartinRistov: moved this within __init__ since it was cluttering up the logs when under _get_client()

PR #248, opened 2024-09-10
botocore is missing in requirements of python-fullstack example

rohit-metrum-ai:
botocore is missing in requirements of python-fullstack example added latest version of botocore==1.35.15 in requirements

### DIFF
--- a/examples/python_fullstack/llama_deploy_app/control_plane/requirements.txt
+++ b/examples/python_fullstack/llama_deploy_app/control_plane/requirements.txt
@@ -1,1 +1,2 @@
 llama-deploy>=0.1.0b20
+botocore==1.35.15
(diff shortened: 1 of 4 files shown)


logan-markewich:
@Rohith-Scalers which package is requiring this? Do you have an example of a traceback?

I was able to run this just fine with the existing requirements 👀

mmezaj:
@logan-markewich i run into the same issue as @Rohith-Scalers. The issues happens in the build of the frontend/dockerfile. 
Docker_Desktop_vtOJQk2QIA

nerdai:
This was my bad @logan-markewich. When testing the `AWSMessageQueue` with our collaborator in that PR, I had to publish a pre-release `v0.1.2ax`. Looks like pip is trying to install this pre release unfortunately.

logan-markewich:
@nerdai can we make one more release to patch the dependencies / imports for `llama_deploy` properly?

nerdai:
Yeah good call. Will try that now.

nerdai:
> @nerdai can we make one more release to patch the dependencies / imports for `llama_deploy` properly?

Alright I published a new patch/pre-release that delays the imports of artifacts from optional deps (namely `botocore`). Everything should work as before now.

logan-markewich:
@Rohith-Scalers @mmezaj your issue should be resolved now

PR #249, opened 2024-09-10
Make services process things faster

### LINKED ISSUE #42: More efficient task processing
Right now, the agent and tools services process one task at a time

This means the processing loop is blocked by the time it takes to execute the single step of task

We can probably make this more efficient 

For agents
- processing a step in the background (i.e. create_task() ?)
- add a callback to handle the case where a task is done and a response should be published

For tools
- most tools probably won't be async
- should we be threading? Or take the same approach as above

logan-markewich:
Currently, the `WorkflowService` is async, but still processes incoming tasks sequentially

We can greatly improve throughput by taking more advantage of `asyncio.create_task` for each incoming task

This PR refactors the `processing_loop()` of the workflow service to be a combination of a task manager and task processor

It can be argued that this should be in the base class, but we haven't cleaned up existing services yet, and I don't want to break them 😅 I could be convinced to do that in this PR though

Fixes https://github.com/run-llama/llama_deploy/issues/42

### DIFF
--- a/llama_deploy/services/workflow.py
+++ b/llama_deploy/services/workflow.py
@@ -46,6 +46,7 @@ class WorkflowServiceConfig(BaseSettings):
     description: str = "A service that wraps a llama-index workflow."
     running: bool = True
     step_interval: float = 0.1
+    max_concurrent_tasks: int = 8
     raise_exceptions: bool = False
 
 
@@ -82,6 +83,7 @@ class WorkflowService(BaseService):
         description (str): The description of the service.
         running (bool): Whether the service is running.
         step_interval (float): The interval in seconds to poll for tool call results. Defaults to 0.1s.
+        max_concurrent_tasks (int): The number of tasks that the service can process at a given time.
         host (Optional[str]): The host of the service.
         port (Optional[int]): The port of the service.
         raise_exceptions (bool): Whether to raise exceptions.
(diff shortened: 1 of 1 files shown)


llama_deploy/services/workflow.py
```diff
@@ -228,78 +234,103 @@ def dump_workflow_state(
             hash=context_hash, state=context_str, run_kwargs=run_kawrgs
         )
 
-    async def processing_loop(self) -> None:
-        """The processing loop for the service.
+    async def process_call(self, task_id: str, current_call: WorkflowState) -> None:
```
nerdai, line 237: nit: do we want a doc string on this?

llama_deploy/services/workflow.py
```diff
@@ -228,78 +234,103 @@ def dump_workflow_state(
...
+                    ).model_dump(),
                 )
+            )
+        finally:
+            # clean up
+            async with self.lock:
+                self._outstanding_calls.pop(task_id, None)
+            self._ongoing_tasks.pop(task_id, None)
+
+    async def manage_tasks(self) -> None:
```
nerdai, line 299: nit: add doc string?

nerdai (approved):
Looks pretty good to me! Left only a couple of nit suggestions

llama_deploy/services/workflow.py
logan-markewich: yes! ha great reminder -- need to keep up the documentation as we go

PR #254, opened 2024-09-13
[Feature] Add aws to `llamaindex/llama-deploy` docker image.

nerdai:
This PR adds the new `AWSMessageQueue` as one of the message queue options in our ready-made docker image.

How was it tested:
- I baked the docker image locally
- Modified the fullstack example to make use of the `awssqs` docker-compose.yml and tested the example still works

image

image

image

### DIFF
--- a/docker/docker-compose-awssqs.yml
+++ b/docker/docker-compose-awssqs.yml
@@ -1,6 +1,6 @@
 services:
   control_plane:
-    image: llamaindex/llama-deploy:local
+    image: llamaindex/llama-deploy:0.1.3
     environment:
       CONTROL_PLANE_HOST: control_plane
       CONTROL_PLANE_PORT: 8000
--- a/docker/docker-compose-awssqs.yml
+++ b/docker/docker-compose-awssqs.yml
@@ -1,0 +1,23 @@
+services:
+  control_plane:
+    image: llamaindex/llama-deploy:local
+    environment:
+      CONTROL_PLANE_HOST: control_plane
+      CONTROL_PLANE_PORT: 8000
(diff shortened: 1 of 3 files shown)


docker/docker-compose-awssqs.yml
```diff
@@ -0,0 +1,23 @@
+services:
+  control_plane:
+    image: llamaindex/llama-deploy:local
```
logan-markewich: Shouldn't this point to our docker hub image? `llamaindex/llama-deploy:0.1.3` for example?

nerdai: Ah yes thank you. I did this for local testing.

Switching it back now.

docker/docker-compose-awssqs.yml
```diff
@@ -0,0 +1,23 @@
...
+      CONTROL_PLANE_INTERNAL_PORT: 8000
+      MESSAGE_QUEUE_CONFIG: awssqs
+      AWS_ACCESS_KEY_ID: $AWS_ACCESS_KEY_ID
+      AWS_SECRET_ACCESS_KEY: $AWS_SECRET_ACCESS_KEY
+      AWS_REGION: $AWS_REGION
+      RUN_MESSAGE_QUEUE: false
+    ports:
+      - "8000:8000"
+    volumes:
+      - ~/.aws/credentials:/root/.aws/credentials:ro # load creds from local
```
nerdai, line 17: Not sure which source of credentials we should use for this docker compose. This method requires the `aws cli` to be configured on the users' local machine.

The alternative seen in lines 11-13 make use of environment variables.

logan-markewich (approved):
LGTM

PR #259, opened 2024-09-18
feat: apiserver config parser

masci:
First step of the apiserver feature introducing the deployment config parser. Notes for the review:
- I ended up not using Pydantic Setting, sparing dozens of lines of code. Since we only use one source to populate the configuration model (the yaml file), the code needed to explain it to Pydantic Settings is way too much.
- I wrote Python 3.11 compatible code: since Llama Deploy is an application and not a library, we should take advantage of the latest features to write better code.
- I slightly adjusted the pre-commit hook, it requires a bit more love but that would be too much to include in this PR


.pre-commit-config.yaml
```diff
@@ -66,17 +66,10 @@ repos:
       - id: prettier
         exclude: poetry.lock
   - repo: https://github.com/codespell-project/codespell
-    rev: v2.2.6
+    rev: v2.3.0
     hooks:
       - id: codespell
         additional_dependencies: [tomli]
-        exclude: ^(poetry.lock|examples|example_scripts)
```
logan-markewich, line 73: We will probably run into examples or lock file items that codespell hits 😅 but, fine for now

logan-markewich (approved):

llama_deploy/apiserver/config_parser.py
```diff
@@ -0,0 +1,48 @@
...
+    name: str
+
+
+class Service(BaseModel):
+    """Configuration for a single service."""
+
+    name: str
+    source: ServiceSource | None = None
+    path: str | None = None
+    port: int | None = None
```
logan-markewich, line 31: I guess our code will handle setting up a host name automatically?

masci: Yeah, I was thinking at least for the first iteration the apiserver would provide a single public entrypoint for a deployment (one of the services listed in the config file) and we can manage the hostnames internally. Which makes me think the config file should probably support the definition of such entrypoint, I'll look into that.

PR #260, opened 2024-09-18
Implement event streaming from workflows

logan-markewich:
This PR adds the ability to
a) expose streaming events from a workflow
b) client code to pass these events in a helpful way to the user. Since we can only put serialized items over the wire, we need to json.dumps the pydantic object. On the client side, I added some logic to at least pass it back to the user as a loaded dict, so that they can rehydrate on their end

I also added some new e2e tests for the CICD, so that we have some assurance that things still work in PRs :) 

What this PR does NOT do
- expose streaming for subworkflows -- I actually don't think this is possible, at least with the handful of ways I was trying it

Things I don't like
- so far, you can use this gist to test https://gist.github.com/logan-markewich/02564c6f1d1135ebe00a9c63ab2da612


llama_deploy/client/async_client.py
```diff
@@ -45,6 +46,17 @@ async def _get_result() -> str:
 
         return await asyncio.wait_for(_get_result(), timeout=self.timeout)
 
+    async def stream_run(
```
logan-markewich: debatable if we should be trying to mimic the workflow interface in the client or not (currently I'm not)

llama_deploy/services/workflow.py
```diff
@@ -248,33 +240,83 @@ async def process_call(self, task_id: str, current_call: WorkflowState) -> None:
...
-            self.load_workflow_state(self.workflow, current_call)
+            ctx = self.load_workflow_state(current_call)
 
             # run the workflow
-            result = await self.workflow.run(**current_call.run_kwargs)
+            handler = self.workflow.run(ctx=ctx, **current_call.run_kwargs)
+            result = await handler
+
+            index = 0
+            if self.streaming:
```
logan-markewich: Was debating not giving the option to enable streaming, and just always have it enabled, but the network overhead might be large

llama_deploy/client/async_client.py
```diff
@@ -104,6 +116,53 @@ async def get_task_result(self, task_id: str) -> Optional[TaskResult]:
...
+                            and isinstance(json_line["result"], str)
+                        ):
+                            try:
+                                yield json.loads(json_line["result"])
+                            except json.JSONDecodeError:
+                                yield json_line["result"]
+                        elif isinstance(json_line, dict) and "result" in json_line:
+                            yield json_line["result"]
+                        else:
+                            yield json_line
```
nerdai: Tight just be me, but there's a bunch of cases here where we yield different things. Would it be possible to add a brief comment with an example case if "json_line" looks like this then we should yield this. Maybe it will make it easier for future us to maintain/update this code if needed?

llama_deploy/client/sync_client.py
```diff
@@ -104,6 +114,58 @@ def get_task_result(self, task_id: str) -> Optional[TaskResult]:
...
+                            and isinstance(json_line["result"], str)
+                        ):
+                            try:
+                                yield json.loads(json_line["result"])
+                            except json.JSONDecodeError:
+                                yield json_line["result"]
+                        elif isinstance(json_line, dict) and "result" in json_line:
+                            yield json_line["result"]
+                        else:
+                            yield json_line
```
nerdai: similar comment left in `async_client`. Probably can just re-use the comments in both places.

llama_deploy/services/workflow.py
```diff
@@ -248,33 +240,83 @@ async def process_call(self, task_id: str, current_call: WorkflowState) -> None:
         """
         try:
             # load the state
-            self.load_workflow_state(self.workflow, current_call)
+            ctx = self.load_workflow_state(current_call)
```
nerdai, line 250: I think for now this will yield `ctx=None` until we implement `load_workflow_state`. Just noting to confirm this should be like this at this point.

llama_deploy/services/workflow.py
nerdai: Maybe we change this in the future. Probably could use some examples of this to assess the potential overhead.

pyproject.toml
```diff
@@ -31,6 +31,7 @@ aio-pika = {version = "^9.4.2", optional = true}
 kafka-python-ng = {version = "^2.2.2", optional = true}
 redis = {version = "^5.0.7", optional = true}
 types-aiobotocore = {version = "^2.14.0", optional = true, extras = ["sqs", "sns"]}
+websockets = ">=9.0.0"
```
nerdai: is this needed?

nerdai (commented):
Generally looks good. I guess it's difficult to test this, should we include an example in our `examples` folder that demonstrates this streaming feature?

llama_deploy/client/async_client.py
logan-markewich: So this can be 
- a `TaskResult` object containing the serialized `ev` from the workflow stream. Since we can rehydrate ourselves, we at least pass back the dict
- a `TaskResult` containing the actual output of the `StopEvent`
- ideally the else never gets hit, but put it there for posterity

llama_deploy/services/workflow.py
logan-markewich: Yea it will always be none -- I actually haven't figured out the UX for managing the workflow state/context yet, so this is mostly dead code at the moment 😅 (I meant to pickle the context here, but you actually can't pickle it do to some asyncio stuff inside it)

pyproject.toml
logan-markewich: no, will remove

.github/workflows/e2e_test.yml
```diff
@@ -0,0 +1,35 @@
...
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install deps
+        shell: bash
+        run: pip install -e .
+      - name: Run All E2E Tests
+        env:
+          CI: true
+        shell: bash
+        run: cd e2e_tests && sh run_all_e2e_tests.sh
```
masci:
```suggestion
        working-directory: e2e_tests
        run: sh run_all_e2e_tests.sh
```

masci (commented):
I'm still relatively new to the codebase so I can't tell much about the implementation details. Some design questions I have: 
- why is streaming tied to a task result? Can't we replicate what we have in workflows, where streamed events and result are independent? The new endpoint `/sessions/{session_id}/tasks/{task_id}/result_stream` could just be `/stream` and produce (serializable) events. The concept of "last result" looks complex to manage, if we can get rid of it the code would be a lot simpler
- why streaming can be optional? It's already opt-in, meaning if no client makes requests to the streaming endpoint nothing will be streamed, and if no service streams events no message will be sent

llama_deploy/client/async_client.py
logan-markewich: Greatly simplified this actually -- it was always be a TaskStream object, with some arbitrary dict as data

llama_deploy/services/workflow.py
logan-markewich: decided to always enable streaming

PR #264, opened 2024-09-20
feat: deployment runtime

masci:
This PR implements the concept of "deployment" that was introduced along with the "apiserver" in #259 

Overview of the abstractions introduced:
- `Deployment`: runs the control plane, the message queue and spawns the services defined in the configuration file
- `Manager`: orchestrate multiple deployment instances
- `SourceManager`: a family of components responsible for managing sources (docker/git) for each deployment
- `GitSourceManager`: a concrete implementation of `SourceManager` specialized for dealing with git sources

Known limitations:
- See the `TODO` and `FIXME` in code comments for the details
- The message queue type is hardcoded, will be made configurable eventually
- Several places in the code assume that all the services are "workflow services"

Next steps:
- Create a thread pool to manage all the concurrent threads runnning different deployments
- Introduce a HTTP-based API for the manager to expose its Python api to clients and CLIs

### DIFF
--- a/llama_deploy/apiserver/deployment.py
+++ b/llama_deploy/apiserver/deployment.py
@@ -27,54 +27,17 @@ class Deployment:
         self._name = config.name
         self._path = root_path / config.name
         self._thread: threading.Thread | None = None
-        self._workflow_services: list[WorkflowService] = []
-
         self._queue = SimpleMessageQueue(**SimpleMessageQueueConfig().model_dump())
         self._control_plane = ControlPlaneServer(
             self._queue.client,
             SimpleOrchestrator(**SimpleOrchestratorConfig().model_dump()),
             **config.control_plane.model_dump(),
         )
+        self._workflow_services: list[WorkflowService] = self._load_services(config)
 
-        for service_id, service_config in config.services.items():
-            source = service_config.source
-            if source is None:
-                # this is a default service, skip for now
-                # TODO: check the service name is valid and supported
-                # TODO: possibly start the default service if not running already
-                continue
-
-            # Sync the service source
-            destination = self._path / service_id
-            source_manager = SOURCE_MANAGERS[source.type]
-            source_manager.sync(source.name, str(destination.resolve()))
-
-            # FIXME: Momentarily assuming everything is a workflow
-            if service_config.path is None:
-                msg = "path field in service definition must be set"
-                raise ValueError(msg)
-
-            # Search for a workflow instance in the service path
-            pythonpath = (destination / service_config.path).parent.resolve()
-            print(pythonpath)
-            sys.path.append(str(pythonpath))
-            module_name, workflow_name = Path(service_config.path).name.split(":")
-            module = importlib.import_module(module_name)
-            workflow = getattr(module, workflow_name)
-            workflow_config = WorkflowServiceConfig(
-                host="workflow",
-                port=8002,
-                internal_host="0.0.0.0",
-                internal_port=8002,
-                service_name=workflow_name,
-            )
-            self._workflow_services.append(
-                WorkflowService(
-                    workflow=workflow,
-                    message_queue=self._queue.client,
-                    **workflow_config.model_dump(),
-                )
-            )
+    @property
+    def name(self) -> str:
+        return self._name
 
     @property
     def path(self) -> Path:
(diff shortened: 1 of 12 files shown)


llama_deploy/apiserver/deployment.py
```diff
@@ -0,0 +1,183 @@
...
+        return self._thread
+
+    def start(self) -> threading.Thread:
+        """Spawns the thread running the asyncio loop for this deployment"""
+        self._thread = threading.Thread(target=asyncio.run, args=(self._start(),))
+        self._thread.start()
+        return self._thread
+
+    async def _start(self) -> None:
+        """The task that will be launched in this deployment asyncio loop.
```
logan-markewich, line 72: So just to confirm, all the message queue, control plane, consumers, and services, are running in the same async loop?

logan-markewich: How would we move that to running things in individual docker containers? Would we need a `DockerDeployment` class eventually?

llama_deploy/apiserver/deployment.py
```diff
@@ -0,0 +1,183 @@
...
+    def _load_services(self, config: Config) -> list[WorkflowService]:
+        """Creates WorkflowService instances according to the configuration object"""
+        workflow_services = []
+        for service_id, service_config in config.services.items():
+            source = service_config.source
+            if source is None:
+                # this is a default service, skip for now
+                # TODO: check the service name is valid and supported
+                # TODO: possibly start the default service if not running already
+                continue
```
logan-markewich, line 111: Should maybe print a warning or raise an error here if the service is unknown?

llama_deploy/apiserver/deployment.py
```diff
@@ -0,0 +1,183 @@
...
+
+            # Search for a workflow instance in the service path
+            pythonpath = (destination / service_config.path).parent.resolve()
+            sys.path.append(str(pythonpath))
+            module_name, workflow_name = Path(service_config.path).name.split(":")
+            module = importlib.import_module(module_name)
+            workflow = getattr(module, workflow_name)
+            workflow_config = WorkflowServiceConfig(
+                host="workflow",
+                port=8002,
```
logan-markewich, line 131: Should the port be grabbed from the config instead?

llama_deploy/apiserver/source_managers/__init__.py
```diff
@@ -0,0 +1,18 @@
+from typing import Protocol
+
+from .git import GitSourceManager
+
+__all__ = ["GitSourceManager"]
+
+
+class SourceManager(Protocol):
```
logan-markewich, line 8: Just curious, why protocol and not ABC ?

llama_deploy/apiserver/deployment.py
masci: Yes, hopefully that will just be a thin layer on top of a real orchestrator like kubernetes, but that's the idea. The current approach is designed to run the whole system in a single process so it can be containerized.

llama_deploy/apiserver/deployment.py
masci: I plan to iterate fast here (I already have the next step ready in a branch), I'm trading unstable code for smaller PRs to ease the review process.

llama_deploy/apiserver/deployment.py
masci: Yes, I'll do it in a separate PR after I finish the deployments orchestration

llama_deploy/apiserver/source_managers/__init__.py
masci: Main reason is I don't expect any need for inheritance, but in general I like duck typing better than anything else in Python, ABC took away all the fun 😅

llama_deploy/apiserver/deployment.py
logan-markewich: Yea makes sense. The reason I asked is running everything in the same deployment introduces some limitations on concurrency vs. services in their own docker containers

llama_deploy/apiserver/source_managers/__init__.py
logan-markewich: Fair!

logan-markewich (approved):

llama_deploy/apiserver/deployment.py
nerdai: If we want to connect a service to a `Deployment` then I think we would just need the `ControlPlaneConfig` in order to register to the `ControlPlane` and to the `MessageQueue` to receive jobs. (At least thats how we've been working up to now.)

So, this initial container can contain `ControlPlane`, `MessageQueue` and optionally any services?

llama_deploy/apiserver/deployment.py
```diff
@@ -0,0 +1,183 @@
...
+        """The task that will be launched in this deployment asyncio loop.
+
+        This task is responsible for launching asyncio tasks for the core components and the services.
+        All the tasks are gathered before returning.
+        """
+        tasks = []
+        # Core components
+        tasks.append(asyncio.create_task(self._queue.launch_server()))
+        # the other components need the queue to run in order to start, give the queue some time to start
+        # FIXME: having to await a magic number of seconds is very brittle, we should rethink the bootstrap process
```
nerdai, line 81: Should we code up a healthcheck here or something similar instead of waiting an arbitrary amount of secs?

nerdai: (future work not for this PR)

nerdai (approved):
Looks like a good start to me!

Minor nit: in the doc strings, I think the first sentence should end with a period ie. "." — I think some doc strings do, but not all of them.

llama_deploy/apiserver/deployment.py
masci: I would add some exponential backoff retry system internally to the various components

PR #266, opened 2024-09-23
feat: Allow configuration of the message queue in the deployment config file

masci:
Add a section in the deployment config file to define the message queue to use.

Notes:
- a string literal field `type` was added to the existing config models so that we can use it to tell Pydantic what's the right model to use when loading `message-queue` from the config

The format will be:
```yaml
message-queue:
  type: simple
  # this section depends on the type of the message queue and will be validated accordingly
  host: "127.0.0.1"
  port: 8001
```

### DIFF
--- a/llama_deploy/apiserver/config_parser.py
+++ b/llama_deploy/apiserver/config_parser.py
@@ -1,6 +1,6 @@
 from enum import Enum
 from pathlib import Path
-from typing import Self, TypeAlias
+from typing import Self, Annotated, Union, Literal
 
 import yaml
 from pydantic import BaseModel, Field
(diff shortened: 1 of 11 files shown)


llama_deploy/apiserver/config_parser.py
```diff
@@ -1,15 +1,59 @@
...
+
+
+class MessageQueueConfigRedis(BaseModel):
+    queue_type: Literal["redis"] = Field("redis", alias="queue-type")
+    config: RedisMessageQueueConfig
+
+
+class MessageQueueConfigSimple(BaseModel):
+    queue_type: Literal["simple"] = Field("simple", alias="queue-type")
+    config: SimpleMessageQueueConfig
```
logan-markewich: It's a little awkward to have `MessageQueueConfigSimple` and `SimpleMessageQueueConfig` -- I wonder if we should just merge these classes by adding queue type to the existing config? Or is it worth maintaining a specific version for yaml config?

logan-markewich (approved):
Looks pretty good to me! Just one comment about some duplicate-ish config objects

llama_deploy/apiserver/config_parser.py
masci: I didn't want to touch the underlying code but indeed we can add a string literal field to the existing config models and get rid of this code

logan-markewich (approved):

PR #267, opened 2024-09-24
feat: add HTTP api to interact with Manager

masci:
This PR adds an HTTP API implemented with FastAPI in front of the deployment Manager.

To run the server, from the llama_deploy folder:
```
python -m apiserver
```

To create a deployment:
```
curl -X POST -F "config_file=@tests/apiserver/data/git_service.yaml" http://localhost:4501/deployments/create/
```


llama_deploy/apiserver/routers/deploy.py
```diff
@@ -0,0 +1,45 @@
+from fastapi import APIRouter, File, UploadFile, HTTPException
```
masci, line 1: I didn't spend too much time for these three endpoints, will define the appropriate response objects later

logan-markewich (approved):
This looks good to me! Very straightforward

A few things are hardcoded (the api server host/port, deployment filepaths), but I'm guessing thats because we still lack the actual CLI for now 👍🏻

masci:
@logan-markewich correct, we need a proper configuration system for the apiserver, likely just its own `PydanticConfig` object.

llama_deploy/apiserver/deployment.py
```diff
@@ -79,7 +79,9 @@ async def start(self) -> None:
         # Spawn SimpleMessageQueue if needed
         if self._simple_message_queue_task:
             # If SimpleMessageQueue was selected in the config file we take care of running the task
-            tasks.append(asyncio.create_task(self._queue.launch_server()))
+            tasks.append(
+                asyncio.create_task(self._simple_message_queue_task.launch_server())
```
nerdai: How come this is no longer `self._queue`? Maybe dumb question, but why do we have these two? (`simple_message_queue_task` is a bit misleading maybe, cause it is a `SimpleMessageQueue` rather than a task?)

nerdai (approved):
I left one question re: about `self._queue` and `self._simple_message_queue_task`, but other than that, looks good!

llama_deploy/apiserver/deployment.py
masci: The problem is with the simple queue we have 2 entities: `SimpleMessageQueue` and `SimpleRemoteClientMessageQueue` and I call "task" the former since it has a `launch_server` method. `self._queue` holds the client, for consistency with the other queue types. I can probably rename `self._queue` to `self._queue_client` for clarity.

masci: Now that I read the code, probably what we call "message queues" should be "message queue clients", yellow flag is no class reimplements `launch_server` and `launch_local` from the base class except the simple message queue. Also `SimpleRemoteClientMessageQueue` raises on `launch_server` and `launch_local`, we might consider a refactoring of that part?

PR #271, opened 2024-09-26
feat: launch apiserver in `llamaindex/llama-deploy` container.

masci:
Previously the entrypoint of the default container `lamaindex/llama-deploy` was designed to simplify the compose files under `examples`. WIth this PR, we start optimizing the container for deploying the apiserver in a production environment.

Example:
```
$ docker run -p 4501:4501 llamaindex/llama-deploy:local
INFO:     Started server process [1]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://0.0.0.0:4501 (Press CTRL+C to quit)
```

From another shell in the same host:
```
$ python -m llama_deploy.cli status  
Llama Deploy is up and running.

Currently there are no active deployments
```

Note this is not breaking for the examples, that are pulling version `v0.1.3` of the image, and that will keep working. We might still need to adjust them at some point so they can use the new image.


masci:
Merging to unblock myself, I'll adapt existing examples later

PR #275, opened 2024-09-27
feat: expose results through the apiserver

masci:
With this PR the apiserver will expose an endpoint `/{deployment_name}/tasks/create` that will return the task result, creating a session for the user under the hood. The corresponding CLI command was created:
```
$ llamactl run -a data "this is my input for the workflow" -d TestDeployment
Received: this is my input for the workflow
```

## How to test
Build a local docker image using this branch. From the `docker` folder of the repo:
```
LLAMA_DEPLOY_VERSION=massi/endpoint docker buildx bake --no-cache
```
(if you need to change the code, or you want to move faster, run `python -m llama_deploy.apiserver` instead)

Run the docker container:
```
docker run -p 4501:4501 llamaindex/llama-deploy:local
```

In a different shell, from the root of the repo, use the CLI to deploy a workflow:
```
python -m llama_deploy.cli deploy tests/apiserver/data/git_service.yaml
```

Same shell, now run a task
```
python -m llama_deploy.cli run -a data foo -d TestDeployment
```


llama_deploy/apiserver/routers/deploy.py
```diff
@@ -1,45 +0,0 @@
-from fastapi import APIRouter, File, UploadFile, HTTPException
```
masci, line 1: renamed to `deployments.py`

tests/apiserver/routers/test_deploy.py
```diff
@@ -1,40 +0,0 @@
-from pathlib import Path
```
masci, line 1: renamed to `test_deployments.py`

llama_deploy/apiserver/app.py
```diff
@@ -3,14 +3,14 @@
...
 
-from .routers import status_router, deploy_router
+from .routers import status_router, deployments_router
 from .server import lifespan
 
 logger = logging.getLogger("uvicorn.info")
 
 
 app = FastAPI(root_path="/", lifespan=lifespan)
-app.include_router(deploy_router)
```
masci, line 13: renamed for consistency, also renamed the python module files

llama_deploy/apiserver/routers/deployments.py
```diff
@@ -0,0 +1,74 @@
...
+        {
+            f"{deployment_name}": "Up!",
+        }
+    )
+
+
+@deployments_router.post("/{deployment_name}/tasks/create")
+async def create_deployment_task(
+    deployment_name: str, task_definition: TaskDefinition
+) -> JSONResponse:
```
masci, line 41: This is new, to be reviewed

logan-markewich: This would be the public endpoint right?

Up to you, but we could even hide the task definition -- just let the user provide an input payload and optional service name (i really need to rename agent_id to service_name lol)

llama_deploy/apiserver/routers/deployments.py
```diff
@@ -0,0 +1,74 @@
...
+
+    if task_definition.agent_id is None:
+        if deployment.default_service is None:
+            raise HTTPException(
+                status_code=400,
+                detail="Service is None and deployment has no default service",
+            )
+        task_definition.agent_id = deployment.default_service
+
+    session = await deployment.client.create_session()
```
logan-markewich, line 55: We should probably delete the session afterwards? Or maybe not?

llama_deploy/cli/run.py
```diff
@@ -0,0 +1,40 @@
...
+@click.option(
+    "-d", "--deployment", required=True, is_flag=False, help="Deployment name"
+)
+@click.option(
+    "-a",
+    "--arg",
+    multiple=True,
+    is_flag=False,
+    type=(str, str),
+    help="'key value' argument to pass to the task, e.g. '-a age 30'",
```
logan-markewich, line 18: I'm actually confused how this works for complex inputs? How is it deciding between the key and value, if the value also has spaces?

masci: Quotes :)

llama_deploy/apiserver/routers/deployments.py
masci: Since it's created "under the hood", I think we should, good catch!

llama_deploy/apiserver/routers/deployments.py
masci: The idea is to keep the HTTP api close to the underlying machinery of llama_deploy for maximum flexibility (say, usage patterns that we didn't anticipate) and hide stuff in the CLI, but we can revisit the approach

llama_deploy/cli/run.py
masci: https://github.com/run-llama/llama_deploy/pull/275/files#diff-b469612cd51d37b4bd64172dd5b1f05fac82cdf9b3ffc6caab3f1160759b0892R49

tests/apiserver/data/git_service.yaml
```diff
@@ -7,5 +7,5 @@ services:
     name: Test Workflow
     source:
       type: git
-      name: https://github.com/run-llama/llama_deploy.git
+      name: https://github.com/run-llama/llama_deploy.git@massi/endpoint
```
nerdai: i guess we'll have to change or remove `@massi/endpoint` before merging?

nerdai (approved):
looks good to me.

llama_deploy/cli/run.py
logan-markewich: Nice!

logan-markewich (approved):
LGTM!

PR #276, opened 2024-09-27
chore: rename --disable-ssl to --insecure

masci:
Minor change, was mentioned in the comments of another PR.

`llamactl` has an option to disable ssl validation when connecting to the apiserver: `-k` and `--disable-ssl`. Since `-k` was shamelessly stolen from `curl`, let's steal the long options as well, so that curl users won't be surprised.


nerdai (approved):
makes sense (no shame here, my friend!)

PR #278, opened 2024-09-27
ux: wrap httpx to uniform connection errors

masci:
Pretty much any command in the CLI will try to connect to the apiserver. To avoid repeating over and over the try/catch block for `httpx.ConnectionError` and possibly make the error message not consistent across the codebase, we wrap `httpx` into an utility method.

### DIFF
--- a/llama_deploy/cli/utils.py
+++ b/llama_deploy/cli/utils.py
@@ -13,6 +13,6 @@ def request(
     except httpx.ConnectError:
         parsed_url = urlparse(str(url))
         raise click.ClickException(
-            "Llama Deploy is not responding, check the apiserver "
+            "Llama Deploy is not responding, check that the apiserver "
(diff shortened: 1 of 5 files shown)


llama_deploy/cli/utils.py
```diff
@@ -0,0 +1,18 @@
...
+
+def request(
+    method: str, url: str | httpx.URL, *args: Any, **kwargs: Any
+) -> httpx.Response:
+    try:
+        return httpx.request(method, url, *args, **kwargs)
+    except httpx.ConnectError:
+        parsed_url = urlparse(str(url))
+        raise click.ClickException(
+            "Llama Deploy is not responding, check the apiserver "
```
nerdai: vert minor nit: "check that the apiserver "

nerdai (approved):
lgtm

left a minor nit in your error message.

PR #280, opened 2024-09-30
feat: add `--timeout` to the CLI

masci:
Default timeout for `httpx` is 5 seconds, but `llamactl deploy` for complex deployments can last more and let the CLI fail with a timeout error. 

As an alternative, we might have a longer, dedicated timeout for the `deploy` command but whatever value we pick, it will look as a magic number, so I preferred to add a global option and leave the burden to the user

### DIFF
--- a/llama_deploy/cli/__init__.py
+++ b/llama_deploy/cli/__init__.py
@@ -18,9 +18,10 @@ from .status import status
     is_flag=True,
     help="Disable SSL certificate verification",
 )
+@click.option("-t", "--timeout", default=5.0, help="Timeout on apiserver HTTP requests")
 @click.pass_context
-def llamactl(ctx: click.Context, server: str, insecure: bool) -> None:
-    ctx.obj = server, insecure
+def llamactl(ctx: click.Context, server: str, insecure: bool, timeout: float) -> None:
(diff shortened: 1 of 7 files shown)


llama_deploy/cli/__init__.py
```diff
@@ -18,9 +18,10 @@
     is_flag=True,
     help="Disable SSL certificate verification",
 )
+@click.option("-t", "--timeout", default=5.0, help="Timeout on apiserver HTTP requests")
```
logan-markewich: Should the default timeout just be `None` ? 5secs feels pretty small for most workloads, and if we can't guess a decent magic number, maybe `None` is better?

logan-markewich: I don't have a strong opinion on this tbh though

logan-markewich (approved):

llama_deploy/cli/__init__.py
masci: I can disable timeouts by default and leave the `-t` option around (it might be handy for using the CLI in scripts)

logan-markewich: Yea I agree with that, the timeout is still helpful, and it can be there for when you need it

PR #281, opened 2024-09-30
feat: install service dependencies in the apiserver

masci:
When deploying a Service, run `pip install` on the list under `python-dependencies` in the configuration file.

Note that the current implementation of the apiserver is single process, so the dependencies will be installed in the virtual environment of the apiserver itself. This solution is good enough for the POC but will be improved as we get closer to production readiness.

### DIFF
--- a/tests/apiserver/data/python_dependencies.yaml
+++ b/tests/apiserver/data/python_dependencies.yaml
@@ -1,0 +1,18 @@
+name: MyDeployment
+
+control-plane:
+  port: 8000
+
+message-queue:
+  type: simple
+  host: "127.0.0.1"
+  port: 8001
+
+default-service: myworkflow
+
+services:
+  myworkflow:
+    name: My Python Workflow
+    python-dependencies:
+      - "llama-index-core<1"
+      - "llama-index-llms-openai"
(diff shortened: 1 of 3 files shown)


tests/apiserver/data/python_dependencies.yaml
```diff
@@ -0,0 +1,18 @@
...
+  port: 8001
+
+default-service: myworkflow
+
+services:
+  myworkflow:
+    name: My Python Workflow
+    python-dependencies:
+      - "llama-index-core<1"
+      - "llama-index-llms-openai"
```
logan-markewich, line 18: If someone puts a requirements file, that will probably need it's own special handling in the future?

logan-markewich (approved):

tests/apiserver/data/python_dependencies.yaml
masci: Yeah I didn't add it because it's not trivial, for example one question would be "should the entry contain an absolute path to the req file? A path relative to the `service.path` field? Or to the `source` root? We can add it with a dedicated PR 👍

logan-markewich: Fair enough! It is complicated yes haha

PR #282, opened 2024-10-01
feat: add services port management

masci:
Service port was hardcoded, meaning two services couldn't be started in the same deployment. With this PR:
- if the port was specified in the config file, the user is responsible to avoid clashes
- if the port is empty in the config file, the manager will keep track of the last port used and increment the next one

### DIFF
--- a/llama_deploy/apiserver/deployment.py
+++ b/llama_deploy/apiserver/deployment.py
@@ -121,7 +121,13 @@ class Deployment:
     def _load_services(self, config: Config) -> list[WorkflowService]:
         """Creates WorkflowService instances according to the configuration object."""
         workflow_services = []
+        default_port = 8002
         for service_id, service_config in config.services.items():
+            port = service_config.port
+            if not port:
+                port = default_port
+                default_port += 1
+
             source = service_config.source
             if source is None:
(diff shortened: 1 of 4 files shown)


logan-markewich (approved):

llama_deploy/apiserver/deployment.py
```diff
@@ -248,6 +254,14 @@ def deploy(self, config: Config) -> None:
...
+
         deployment = Deployment(config=config, root_path=self._deployments_path)
         self._deployments[config.name] = deployment
         self._pool.apply_async(func=asyncio.run, args=(deployment.start(),))
+
+    def _assign_control_plane_port(self, config: Config) -> None:
+        for service in config.services.values():
+            if not service.port:
+                service.port = self._control_plane_port
+                self._control_plane_port += 1
```
logan-markewich, line 267: I guess this doesn't check for collision on services that have a port specified

masci: It doesn't, no. The design philosophy is: the moment users specify a port, they're responsible for avoiding clashes. This is only ok for the "single process" implementation, as we scale we'll need to revisit this part.

PR #283, opened 2024-10-02
docs: Refactor the Python Fullstack example to run with the apiserver

masci:
This is a rewrite of the fullstack example in order to use the new Llama Deploy apiserver. Notable changes:
- The `workflow` folder is now a Python package, so that the apiserver can import the workflow instances from there
- We don't `include` anymore, all the components are in the same `docker-compose` file for clarity


examples/python_fullstack/frontend/frontend/state.py
```diff
@@ -38,12 +36,21 @@ async def answer(self):
...
-            "agentic_workflow",
-            chat_history_dicts=chat_history_dicts,
-            user_input=question,
+        input_payload = {
+            "chat_history_dicts": chat_history_dicts,
+            "user_input": question,
+        }
+        deployment_name = os.environ.get("DEPLOYMENT_NAME", "MyDeployment")
+        apiserver_url = os.environ.get("APISERVER_URL", "http://localhost:4501")
+        response = await client.post(
```
logan-markewich, line 48: Are we considering an actual client for the api-server? 👀 Making http requests yourself isn't the most user friendly

logan-markewich (approved):
looks pretty good to me, but just a few issues with linting/tests

examples/python_fullstack/frontend/frontend/state.py
masci: I think we should add apiserver support to the existing client, yes.

PR #294, opened 2024-10-06
CORS Configuration and Session ID Default

jonpspri:
Added two bits to Control Plane:

1.  If Session ID is not in a Task Definition, add it based on the session in the original POST URL.  (Avoid having two potentially different sources of truth).

2.  Add configurable (off by default) CORS support.  WIll probably build it out with other CORS configuration features -- I want to be able to support browser-based access at least during development.  LMK if that's a BAD IDEA -- I could always switch to proxy-based CORS and access management.

Cheers!


logan-markewich (approved):
This looks pretty good to me!

I did roll back the change with the context though -- this should be addressed by #306

PR #302, opened 2024-10-08
lazy init the message queue configs

logan-markewich:
In this example, we should lazy-init the message queue configs to avoid requiring certain env variables


nerdai (approved):
oh good catch -- thanks!

logan-markewich:
This + v0.2.1 also fixes https://github.com/run-llama/llama_deploy/issues/284

PR #306, opened 2024-10-09
Re-introduce Proper State Management

logan-markewich:
This PR does a few things
- adds new API's to the control plane to get/set state for a given session
- no longer sends state over the message queue (this was abusive to begin with)
- removes `NewTask` (since the separation between state and task def is no longer needed in the message queue)
- uses proper context serialization for the workflow 
- adds an e2e test to ensure this actually works

A follow-up PR will update our full stack example to include controls for sessions


logan-markewich:
Note that this still relies on the kvstore passed into the control plane

Tbh, I think I want to keep it this way:
- we can default to something more reliable like a duckdb kvstore. However, this will not scale horizontally 
- writing a scalable server for duckdb sounds like a nightmare
- if a user wants something more scalable, they can plug in another kvstore -- I know originally we wanted to keep this internal, but the eng lift to write a distributed kvstore server does not feel worth it :) (I would just use mongodb or redis as the default, but those need external servers running)

masci (approved):
Nothing to say, changes look good to me and the idea explained in the PR body makes sense to me!

nerdai (approved):
lgtm! As follow up, we may want to add the respective session state endpoints to the api server?

logan-markewich:
yes @nerdai that should be the next step (or at a least one of them!) 🙏🏻

PR #313, opened 2024-10-12
Support 3.10

Disiok:


logan-markewich (approved):
Works for me! Assuming cicd passes

Thanks for getting tests to work!

logan-markewich:
(We can excuse the 0.01% coverage drop)

PR #315, opened 2024-10-12
fix cli run command

logan-markewich:
The /task/create endpoint no longer blocks and returns a result, switch the endpoint being called to /task/run


nerdai (approved):
Oops. I should've caught this. Thanks for fixing!

PR #319, opened 2024-10-15
fix: actually stream events from the async client

### LINKED ISSUE #308: All Events in Result Stream Arrive At Once Instead of When Ready
I am using Llama deploy for an application where I want to send a client updates as a workflow is running. I'm trying to do this with the get_task_result_stream function in an AsyncLlamaDeployClient. 

In the E2E test, the workflow has Steps 1 through 3, which produce the events with the numbers 0.3, 0.6 and 0.9. I added a "await asyncio.sleep(2)" after each of the events `ctx.write_event_to_stream(ProgressEvent(progress=0.3))`...etc. 

See below. 

```
class StreamingWorkflow(Workflow):
    @step()
    async def run_step_1(self, ctx: Context, ev: StartEvent) -> Step1:
        arg1 = ev.get("arg1")
        if not arg1:
        ctx.write_event_to_stream(ProgressEvent(progress=0.3))
        await asyncio.sleep(2)
        return Step1(arg1=str(arg1) + "_result")

    @step()
    async def run_step_2(self, ctx: Context, ev: Step1) -> Step2:
        arg1 = ev.arg1
        if not arg1:
         ctx.write_event_to_stream(ProgressEvent(progress=0.6))
        await asyncio.sleep(2)
        return Step2(arg1=str(arg1) + "_result")

    @step()
    async def run_step_3(self, ctx: Context, ev: Step2) -> StopEvent:
        arg1 = ev.arg1
        if not arg1:
        ctx.write_event_to_stream(ProgressEvent(progress=0.9))
        await asyncio.sleep(2)
```

In `test_run_client.py`, I modified it so that the event's progress will be printed out. I'm noticing that these three events (0.3, 0.6, 0.9) are arriving at the same time after 6 seconds. The desired behavior is that each event arrives separately after 2 seconds. 

```
  async for event in session.get_task_result_stream(task_id):
      if "progress" in event:
          logger.info(event["progress"])
```

Any insight into how to solve this would be greatly appreciated!

masci:
Fixes https://github.com/run-llama/llama_deploy/issues/308

Changes:
- Fixed the way `httpx` is used to consume the `StreamResponse` from FastAPI
- Removed some duplicated code (to be double checked during review)


llama_deploy/client/async_client.py
```diff
@@ -126,26 +127,26 @@ async def get_task_result_stream(
...
-                    )
+        while True:
+            try:
+                with httpx.stream(
+                    "GET",
+                    f"{self.control_plane_url}/sessions/{self.session_id}/tasks/{task_id}/result_stream",
+                ) as response:
                     response.raise_for_status()
-                    async for line in response.aiter_lines():
+                    for line in response.iter_lines():
```
logan-markewich: Shouldn't this be async? Now the stream is blocking 👀

llama_deploy/control_plane/server.py
```diff
@@ -549,41 +549,18 @@ async def event_generator(
...
-                if final_result is not None:
-                    return
-
-                # Continue to check for new results
+                last_index = 0
                 while True:
-                    await asyncio.sleep(
-                        self.step_interval
-                    )  # Small delay to prevent tight loop
-                    session = await self.get_session(session_id)
```
logan-markewich, line 575: If we don't refresh the session object, how do we get new stream events on the next loop?

masci: wouldn't `self.get_session(session_id)` always return the same session?

llama_deploy/client/async_client.py
masci: Yep, I was only looking at the server side of things, fixed

llama_deploy/control_plane/server.py
masci: Ok apparently it wouldn't 😅 
Added `session = await self.get_session(session_id)` at every cycle of the while loop

logan-markewich: Yea, since the session is technically stored "off-site" (i.e. in a kvstore), so its not in process/in-memory

logan-markewich (approved):

PR #324, opened 2024-10-16
v0.2.4

logan-markewich:


masci:
Pulled in the lockfile to fix starlette CVE

masci (approved):

masci:
Can we also ship #326 ?

PR #327, opened 2024-10-18
feat: add apiserver support to Python SDK

### LINKED ISSUE #290: Python SDK should support calling the APIServer
Extend `Client` and `AsyncClient` from the Python SDK to support the endpoints exposed by the API Server

masci:
Introducing apiserver support through a new Python SDK.

fixes #290 

## Basic usage
```py
from llama_deploy.client import Client

# Use the same client for async and sync operations
c = Client()

async def an_async_function():
    status = await client.apiserver.status()

def normal_function():
    status = client.sync.apiserver.status()
```

## Streaming events
```py
with open("deployment.yml") as f:
    deployments = await client.apiserver.deployments()
    d = await deployments.create(f)

tasks = await d.tasks()
task = await tasks.create(TaskDefinition(input='{"a": "b"}'))
async for ev in task.events():
    print(ev)
```


e2e_tests/apiserver/test_streaming.py
```diff
@@ -0,0 +1,21 @@
...
+
+    with open(here / "deployments" / "deployment_streaming.yml") as f:
+        deployments = await client.apiserver.deployments()
+        deployment = await deployments.create(f)
+        await asyncio.sleep(5)
+
+    tasks = await deployment.tasks()
+    task = await tasks.create(TaskDefinition(input='{"a": "b"}'))
+    async for ev in task.events():
+        print(ev)
```
logan-markewich: nit: should we assert that we get the expected number of events back, or the contents of the events?

llama_deploy/apiserver/routers/deployments.py
```diff
@@ -144,6 +143,23 @@ async def get_task_result(
...
+    deployment = manager.get_deployment(deployment_name)
+    if deployment is None:
+        raise HTTPException(status_code=404, detail="Deployment not found")
+
+    tasks: list[TaskDefinition] = []
+    for session_def in await deployment.client.list_sessions():
+        session = await deployment.client.get_session(session_id=session_def.session_id)
+        for task_def in await session.get_tasks():
+            tasks.append(task_def)
+    return JSONResponse(tasks)
```
logan-markewich, line 160: Should this be a dict of session_id -> list[task] ? (Just thinking about building a UI that would use this function)

llama_deploy/client/models/model.py
```diff
@@ -0,0 +1,71 @@
...
+        """We prevent the usage of the constructor and force users to call `instance()` instead."""
+        raise TypeError("Please use instance() instead of direct instantiation")
+
+    @classmethod
+    def instance(cls, make_sync: bool = False, **kwargs: Any) -> Self:
+        """Returns an instance of the given model.
+
+        Using the class constructor is not possible because we want to alter the class method to
+        accommodate sync/async usage before creating an instance, and __init__ would be too late.
+        """
```
logan-markewich, line 29: Ha, this docstring answered so many questions. Thank you!

llama_deploy/client/models/model.py
```diff
@@ -0,0 +1,71 @@
...
+
+    class Wrapper(_class):  # type: ignore
+        pass
+
+    for name, method in _class.__dict__.items():
+        # Only wrap async public methods
+        if asyncio.iscoroutinefunction(method) and not name.startswith("_"):
+            setattr(Wrapper, name, async_to_sync(method))
+    # Static type checkers can't assess Wrapper is indeed a type[T], let's promise it is.
+    return cast(type[T], Wrapper)
```
logan-markewich, line 71: So just confirming how this works
- we define a class with all async methods
- if a flag is set, we apply this wrapper
- this wrapped only converts public methods
- type checkers and IDEs will see this as a sync method instead of async?

🤯

llama_deploy/client/__init__.py
logan-markewich, line 1: Can we also export the `Client` at the top level?

llama_deploy/client/base.py
```diff
@@ -0,0 +1,22 @@
...
+
+import httpx
+
+from .client_settings import ClientSettings
+
+
+class _BaseClient:
+    """Base type for clients, to be used in Pydantic models to avoid circular imports."""
+
+    def __init__(self, **kwargs: Any) -> None:
```
logan-markewich: Thoughts on adding typed args?
image

llama_deploy/client/models/model.py
logan-markewich: Hmm, seems like IDE/type-checker assumes its always async (which tbh is fine)

llama_deploy/client/models/apiserver.py
```diff
@@ -0,0 +1,289 @@
...
+DEFAULT_POLL_INTERVAL = 0.5
+
+
+class Session(Model):
+    """A model representing a session."""
+
+    pass
+
+
+class SessionCollection(Collection):
```
logan-markewich, line 21: Is this missing a few methods? I don't see how I can create a new task under a session (just playing around a bit with the code in vscode, trying to understand the usage pattern)

llama_deploy/client/base.py
masci: I merged the settings into the client itself, making this work. If we like accessing settings like `client.apiserver_url`, autocompletion works now.

llama_deploy/client/models/model.py
masci: Ah correct, didn't think about that so to clarify, the static checker would be happy with `await client.sync.apiserver.status()` because the actual wrapping happens at runtime. I agree this would be fine, but I'll think about possible implications

llama_deploy/apiserver/routers/deployments.py
masci: Good point, many return payloads of the apiserver API are not well thought, we can start fixing it from here

llama_deploy/client/models/apiserver.py
masci: We should be good now. Creating a task from a session instance is not part of `apiserver` so it'll be added later when I'll migrate the other "models"

llama_deploy/apiserver/routers/deployments.py
masci: I had a look and I realised if we change this we need to update the CLI as well and the PR would grow considerably. I propose we do that in a follow up.

masci: Tracked in https://github.com/run-llama/llama_deploy/issues/337

logan-markewich: Sounds good to me!

logan-markewich (approved):

PR #331, opened 2024-10-19
chore: make the docker build more customizeable

masci:
While debugging a problem with the docker image I had to bisect-search a branch and I needed to build specific commits and made these changes.

Tought this could be helpful again in the future so we might add it to the codebase.


logan-markewich (approved):
Looks good! Thanks!

Issue #335, opened 2024-10-24
Port existing clients to the new `Client` class


masci:
With #290 we only introduced support for the apiserver API, but `Client` should also support the control plane and in general replicate all that was possible with `LlamaDeployClient`, `SessionClient` and their async counterparts.

PR #340, opened 2024-10-25
feat: Add `services` registration API to `Client`

masci:
Part of #335

- Added `client.core.services`

Also in this PR:
- remove the need to pass `timeout` to the client every single request. It can still be passed if one needs a custom value.
- move common fixtures to global `conftest.py` to reduce duplication
- removed `scope=session` so every e2e function has a fresh control plane to test

### DIFF
--- a/llama_deploy/client/models/core.py
+++ b/llama_deploy/client/models/core.py
@@ -1,0 +1,46 @@
+from llama_deploy.types.core import ServiceDefinition
+
+from .model import Collection, Model
+
+
+class Service(Model):
+    pass
+
+
+class ServiceCollection(Collection):
+    async def register(self, service: ServiceDefinition) -> None:
+        """Registers a service with the control plane.
+
+        Args:
+            service: Definition of the Service to register.
+        """
+        register_url = f"{self.client.control_plane_url}/services/register"
+        await self.client.request("POST", register_url, json=service.model_dump())
+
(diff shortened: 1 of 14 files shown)


llama_deploy/client/models/core.py
```diff
@@ -0,0 +1,54 @@
...
+        deregister_url = f"{self.client.control_plane_url}/services/deregister"
+        await self.client.request(
+            "POST",
+            deregister_url,
+            params={"service_name": service_name},
+        )
+        self.items.pop(service_name)
+
+
+class Core(Model):
```
logan-markewich, line 38: For now, `core` only has `services` -- hypothetically, what else do you think would go here in the future? :eyes:

logan-markewich (approved):
Looks great! Love the test cleanup

llama_deploy/client/models/core.py
masci: Mainly `sessions`. Basically I'm moving under `core` all the client capabilities needed to talk to the ControlPlane.

logan-markewich: gotcha

masci: The hope is we can direct users towards two directions: `apiserver` (managed deployments) and `core` (manual deployments) to map the user stories we have in the docs.

PR #343, opened 2024-10-29
refact: make collections in apiserver models lazy

masci:
Adjust `client.apiserver.deployments` and `client.apiserver.sessions` so that they don't actually load the items in a collection, deferring that to `list`. This way `sessions` and `deployments` can be regular properties, accessible without `await`.
This was originally implemented for the `core` client, this PR makes it consistent with the rest of the models.

Part of #337

### DIFF
--- a/llama_deploy/client/models/apiserver.py
+++ b/llama_deploy/client/models/apiserver.py
@@ -9,8 +9,6 @@ from llama_deploy.types.core import SessionDefinition, TaskDefinition, TaskResul
 
 from .model import Collection, Model
 
-DEFAULT_POLL_INTERVAL = 0.5
-
 
 class Session(Model):
     """A model representing a session."""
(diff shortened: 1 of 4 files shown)


llama_deploy/client/models/apiserver.py
```diff
@@ -150,58 +169,51 @@ async def create(self, task: TaskDefinition) -> Task:
...
-                id=session_def.session_id,
-            )
-            for session_def in r.json()
-        }
         return SessionCollection.instance(
             make_sync=self._instance_is_sync,
             client=self.client,
             deployment_id=self.id,
-            items=items,
+            items={},
```
logan-markewich, line 216: Does a SessionCollection (or other collections) even need an `items` kwarg/attribute on the instance now?

llama_deploy/client/models/apiserver.py
```diff
@@ -150,58 +169,51 @@ async def create(self, task: TaskDefinition) -> Task:
...
-        )
-        items = {
-            "id": Session.instance(
-                make_sync=self._instance_is_sync,
-                client=self.client,
-                id=session_def.session_id,
-            )
-            for session_def in r.json()
-        }
         return SessionCollection.instance(
```
logan-markewich, line 212: This makes sense to me

Conceptually, this is returning a client, and from there, you interact with it

llama_deploy/client/models/apiserver.py
masci: Correct, collections don't need it anymore (unless we want to introduce some sort of caching but at this stage I'd rather not)

logan-markewich (approved):

PR #344, opened 2024-10-31
fix: fix typing for wrapped async methods

masci:
The problem with the previous approach was that the static checker couldn't distinguish between the original async and the sync-wrapped version of methods, leading to typing errors because `client.sync.something()` was not awaited.

The workaround is to tell the static checkers to ignore the types of the sync-wrapped methods marking them with `Any`. This way mypy and pyright won't complain. The drawback is the `client.sync.*` methods won't have autocompletion (copilot and friends would still work tho).

### DIFF
--- a/llama_deploy/client/client.py
+++ b/llama_deploy/client/client.py
@@ -1,5 +1,7 @@
+from typing import Any
+
 from .base import _BaseClient
-from .models import ApiServer, Core
+from .models import ApiServer, Core, make_sync
 
 
 class Client(_BaseClient):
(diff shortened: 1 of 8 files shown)


llama_deploy/client/models/model.py
```diff
@@ -16,24 +17,10 @@ class _Base(BaseModel):
...
-            cls = _make_sync(cls)
-
-        inst = super(_Base, cls).__new__(cls)
-        inst.__init__(**kwargs)  # type: ignore[misc]
-        inst._instance_is_sync = make_sync
-        return inst
+    def _prepare(self, _class: type) -> type:
+        if self._instance_is_sync:
+            return make_sync(_class)
+        return _class
```
masci, line 21: I changed the logic around getting instances according to the sync/async desired status in favor of getting classes, it's much more simple and we get control over the type returned for the classes converted to sync. With `instance()` the type of the instance returned was always the same as the original async type, making static checkers go nuts for the lack of `await` in the caller code using `client.sync.foo`

llama_deploy/client/client.py
```diff
@@ -31,19 +33,19 @@ def sync(self) -> "_SyncClient":
...
-        return Core.instance(client=self, id="core")
+        return Core(client=self, id="core")
 
 
-class _SyncClient(Client):
+class _SyncClient(_BaseClient):
     @property
-    def apiserver(self) -> ApiServer:
-        return ApiServer.instance(make_sync=True, client=self, id="apiserver")
+    def apiserver(self) -> Any:
```
masci, line 46: The `Any` here is basically telling the type checker "mind your own business". This means autocompletion an IDE checkers won't work for `client.sync.*` but it's the best I could achieve. It's honest from the typing point of view (since we change the async nature of methods at runtime) and it's better than producing errors I think.

logan-markewich (approved):
This makes sense to me. Looks a little hairy lol  but I think I understand what's going on here

PR #345, opened 2024-10-31
Integrate the Solace PubSub+ broker into LlamaDeploy

alimosaed:
**What is the purpose of this change?**
This pull request integrates the Solace PubSub+ broker into LlamaDeploy.

**How is this accomplished?**
To complete the integration, the following tasks were performed:

1. Added Solace PubSub+ as a message queue.
2. Included sample workflows to test the integration.
3. Dockerized both the Solace broker and LlamaDeploy for easy deployment.


llama_deploy/message_queues/solace/boot.py
```diff
@@ -0,0 +1,77 @@
+import os
+from logging import getLogger
+from typing import Any, Dict, List, Literal, TYPE_CHECKING
+
+# if TYPE_CHECKING:
+from solace.messaging.config.solace_properties import (
```
logan-markewich: If these aren't under `if TYPE_CHECKING:` blocks, imports will fail

Move these under type checking blocks, and then import them when they are actually needed, and raise an appropriate message if the import fails (by default, these packages will not be installed)

llama_deploy/message_queues/solace/boot.py
```diff
@@ -0,0 +1,77 @@
...
+SEMP_USERNAME_KEY = 'solace.semp.username'
+SEMP_PASSWORD_KEY = 'solace.semp.password'
+SEMP_PORT_TO_CONNECT = '1943'
+VALID_CERTIFICATE_AUTHORITY = 'public_root_ca'
+IS_QUEUE_TEMPORARY = 'IS_QUEUE_TEMPORARY'
+
+# Configure logger
+logger = getLogger(__name__)
+
+class BootSolace:
```
logan-markewich: I don't 100% understand the purpose of this class? Shouldn't this all just be part of the queue config or message queue itself? Seems like its just validating the env?

llama_deploy/message_queues/solace/solace.py
```diff
@@ -0,0 +1,275 @@
...
+
+from pydantic import BaseModel, Field
+from pydantic_settings import BaseSettings, SettingsConfigDict
+
+from llama_deploy.message_queues.base import BaseMessageQueue
+from llama_deploy.messages.base import QueueMessage
+from llama_deploy.message_consumers.base import BaseMessageQueueConsumer, StartConsumingCallable
+
+# if TYPE_CHECKING:
+from solace.messaging.errors.pubsubplus_client_error import PubSubPlusClientError, IllegalStateError
```
logan-markewich: Similar comment here about type checking

llama_deploy/message_queues/solace/solace.py
```diff
@@ -0,0 +1,275 @@
...
+            # Process the message using the consumer
+            asyncio.run(self._consumer.process_message(queue_message))
+
+            if self._receiver:
+                self._receiver.ack(message)
+                
+        except Exception as unexpected_error:
+            logger.error(f"Error consuming message: {unexpected_error}")
+
+class SolaceMessageQueueConfig(BaseSettings):
```
logan-markewich: There's nothing in this config class? Generally everything will be in the config here, and then the message queue just uses the config to instantiate. Seems like a loot of the boot file should be in the config?

llama_deploy/message_queues/solace/boot.py
alimosaed: Thanks for your comment. I’ve moved some packages under type checking.

llama_deploy/message_queues/solace/boot.py
alimosaed: This class was responsible to get the required Solace broker's configurations from the environment variables. I moved this script's functionalities to the Solace.py file and simplified its logic.

llama_deploy/message_queues/solace/solace.py
alimosaed: I’ve moved one package under type checking. 

I tried to move PersistentMessageReceiver and PersistentMessagePublisher as well but it caused an error, so I had to import them instead. The remaining imported packages are not used for the type checking.

llama_deploy/message_queues/solace/solace.py
alimosaed: I updated the class and added the configuration logic.

alimosaed (commented):
Thank you @logan-markewich for the comments. I addressed them. I appreciate it if you review the PR again and give me your feedback.

llama_deploy/message_queues/solace/solace.py
logan-markewich: These types cannot be imported at the top-level. Since solace is in the `__init__.py` file, it will trigger the imports from this file, and will raise errors if the user doesn't have solace installed.

Check out some of the other message queues. You should be able to import these in the functions that use them. If thats absolutely not possible, remove it from the shortcut imports in the `__init__.py` file

llama_deploy/message_queues/solace/solace.py
logan-markewich: With pydantic settings, it will automatically read from env vars

You can modify the class like

```python
class SolaceMessageQueueConfig(BaseSettings):
  model_config = SettingsConfigDict(env_prefix="SOLACE_")
  type: Literal["solace"] = Field(default="solace", exclude=True)
  host: str = Field(default="")
  vpn_name: str = Field(default="")
  ...
```

And so on. This way, the user can set `SOLACE_HOST` and this settings object will automatically pick it up. This settings object is also used to pass in settings through the yml config instead of env vars, which is why its important to have actual fields/settings on this config object

llama_deploy/message_queues/solace/solace.py
alimosaed: I applied a few changes:
- I removed the Solace folder and moved the solace.py to the parent folder where other brokers exist.
- I could not move all imports to the classes or methods. So, I had to remove Solace from the shortcut imports.

llama_deploy/message_queues/solace/solace.py
alimosaed: I replaced the dotenv with the pydantic setting.

alimosaed (commented):
I addressed your comments.

Thanks.

examples/message-queue-integrations/_app/pyproject.toml
```diff
@@ -14,4 +14,4 @@ python = "^3.10"
 llama-index-llms-openai = "^0.2.1"
 llama-index-embeddings-openai = "^0.2.4"
 llama-index-agent-openai = "^0.3.0"
-llama-deploy = {version = "^0.2.1", extras = ["awssqs", "rabbitmq", "kafka", "redis"]}
+llama-deploy = {version = "^0.2.1", extras = ["awssqs", "rabbitmq", "kafka", "redis", "solace"]}
```
logan-markewich: probably should be `^0.3.0`

logan-markewich:
@alimosaed Did you run `make lint` ?  Linting is failing 
 
Also, you'll need to run `poetry lock` and update the lock file, e2e tests are complaining that its out of date

If you can add any sort of test to help the code coverage not decrease so much, that would be nice too

alimosaed:
Thanks @logan-markewich 
I fixed the lint errors and ensured that make test and lint are successfully passed. I will add some unit tests.

examples/message-queue-integrations/_app/pyproject.toml
alimosaed: Updated the this version and it worked this time.

alimosaed:
@logan-markewich I added some unit tests. It is ready for the review. Thanks.

logan-markewich:
Looks like its nearly ready, but some merge conflicts (maybe with the poetry.lock file? You might have to merge main and run lock again)

alimosaed:
Thanks @logan-markewich . I resolved conflicts.

logan-markewich (approved):

PR #351, opened 2024-11-06
feat: add get_task_result_stream to new client

### LINKED ISSUE #335: Port existing clients to the new `Client` class

masci:
This PR ports the last method missing from `AsyncSessionClient`, `get_task_result_stream`.

Fixes https://github.com/run-llama/llama_deploy/issues/335

Notes for the reviewer:
- async methods returning `Coroutine[AsyncGenerator]` when converted with `async_to_sync` return `AsyncGenerator`, that we can't use in sync calls. To overcome this limitation, `generator_wrapper` was added that consumes the async generator under the hood and returns a list
- end-to-end tests were update to use `Client`
- unit tests were added to cover the new wrapper

### DIFF
--- a/llama_deploy/client/client.py
+++ b/llama_deploy/client/client.py
@@ -1,3 +1,4 @@
+import asyncio
 from typing import Any
 
 from .base import _BaseClient
@@ -28,7 +29,13 @@ class Client(_BaseClient):
     @property
     def sync(self) -> "_SyncClient":
         """Returns the sync version of the client API."""
-        return _SyncClient(**self.model_dump())
+        try:
+            asyncio.get_running_loop()
+        except RuntimeError:
+            return _SyncClient(**self.model_dump())
+
+        msg = "You cannot use the sync client within an async event loop - just await the async methods directly."
+        raise RuntimeError(msg)
 
     @property
     def apiserver(self) -> ApiServer:
--- a/llama_deploy/client/models/model.py
+++ b/llama_deploy/client/models/model.py
@@ -1,8 +1,10 @@
 import asyncio
-from typing import Any, Generic, TypeVar
+import inspect
+from typing import Any, AsyncGenerator, Callable, Generic, TypeVar
 
 from asgiref.sync import async_to_sync
 from pydantic import BaseModel, ConfigDict, Field, PrivateAttr
+from typing_extensions import ParamSpec
 
 from llama_deploy.client.base import _BaseClient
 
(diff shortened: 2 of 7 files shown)


llama_deploy/client/models/model.py
```diff
@@ -42,15 +44,37 @@ def list(self) -> list[T]:
...
 
-    class Wrapper(_class):  # type: ignore
+    class ModelWrapper(_class):  # type: ignore
         _instance_is_sync: bool = True
 
+    def generator_wrapper(
+        func: Callable[_P, AsyncGenerator[_G, None]], /, *args: Any, **kwargs: Any
+    ) -> Callable[_P, list[_G]]:
+        def new_func(*fargs: Any, **fkwargs: Any) -> list[_G]:
+            return asyncio.run(_async_gen_to_list(func(*fargs, **fkwargs)))
```
logan-markewich, line 69: If there's already an event loop, this will fail without nest_asyncio

logan-markewich (approved):
Lgtm honestly, nice

llama_deploy/client/models/model.py
masci: Good catch. This line in particular is not even the problem, `async_to_sync` would raise a runtime error if you try to use the converted sync code in an async function within an existing loop (fair enough, why would you want all the overhead?). I added a [check](https://github.com/run-llama/llama_deploy/pull/351/files#diff-b9a1ab538251b8ead629a4d316eeed3f63293af89c09cf5778672a53399c3ad8R38) so it's our client raising `RuntimeError` now with a specific message explaining what's happening.

llama_deploy/client/client.py
```diff
@@ -28,7 +29,13 @@ def normal_function():
     @property
     def sync(self) -> "_SyncClient":
         """Returns the sync version of the client API."""
-        return _SyncClient(**self.model_dump())
+        try:
+            asyncio.get_running_loop()
+        except RuntimeError:
+            return _SyncClient(**self.model_dump())
+
+        msg = "You cannot use the sync client within an async event loop - just await the async methods directly."
```
nerdai, line 37: Oh, was this something we were running into before. Anyways, nice catch!

llama_deploy/client/models/model.py
```diff
@@ -42,15 +44,37 @@ def list(self) -> list[T]:
         return [self.get(id) for id in self.items.keys()]
 
 
+# Generic type for what's returned by the async generator
+_G = TypeVar("_G")
+# Generic parameter for the wrapped generator method
+_P = ParamSpec("_P")
+# Generic parameter for the wrapped generator method return value
+_R = TypeVar("_R")
```
nerdai, line 52: Possibly dumb question as perhaps I just missed this, but where are we using this generic param?

nerdai (approved):
LGTM! Just had one question on `_R` generic param that was introduced and I couldn't see where it was being used?

llama_deploy/client/models/model.py
logan-markewich: Nice!

llama_deploy/client/models/model.py
masci: Not dumb, leftover! I'll remove it, good catch!

llama_deploy/client/client.py
masci: This problem always existed but didn't run into it until @logan-markewich made me notice!

PR #353, opened 2024-11-07
feat: make the kafka topic configurable

masci:
Part of #349 (not a fix until we propagate the topic settings up to the control plane and api server)

## Problem

Currently we use the name of the "message type" string to create the Kafka topic that'll be used in the message queue. Problem is that different instances of control planes using the same Kafka instances will step into each other, with catastrific effects. Note that this problem is common to other message queue brokers like S3, so eventually this fix will be rolled out for all the supported queues.

## Solution

Make the Kafka topic name configurable, so that different control planes will use different topics. This PR only makes the topic configurable, another PR will make use of this new feature and resolve the original issue.

## Notes

While working at the fix, I noticed a few things we can improve and already implemented in this PR in a backward compatible manner:
- I am not sure the base class for queues should be a pydantic model, that's very limiting when you need to treat the class as a regular class (get and set local instance variables for example). I think a plain abstract class derived from `ABC` will make the job.
- There's a lot of duplication in the configuration settings: the class constructor, the config settings model, the static methods... I think we should have a single source of truth, being the config settings model. For Kafka I made the constructor take the config model instance directly.
- I moved the original code behind the `if __name__ == "__main__"` that was used for testing into a proper e2e test. Tests for message queues will most certainly need docker to be avaiable, I'll add the required CI bits (this part is WIP, you'll see it's done when tests are all green).


llama_deploy/apiserver/deployment.py
```diff
@@ -209,7 +209,7 @@ def _load_message_queue_client(
         if cfg.type == "aws":
             return AWSMessageQueue(**cfg.model_dump())
         elif cfg.type == "kafka":
-            return KafkaMessageQueue(**cfg.model_dump())
+            return KafkaMessageQueue(**cfg.model_dump())  # type: ignore
```
masci: This is temporarily needed, until all the queues derive from `AbstractMessageQueue` and we can change the return type of the function accordingly.

llama_deploy/deploy/deploy.py
```diff
@@ -67,7 +67,7 @@ def _get_message_queue_client(config: BaseSettings) -> BaseMessageQueue:
     elif isinstance(config, AWSMessageQueueConfig):
         return AWSMessageQueue(**config.model_dump())
     elif isinstance(config, KafkaMessageQueueConfig):
-        return KafkaMessageQueue(
+        return KafkaMessageQueue(  # type: ignore
```
masci: temporary, see above

llama_deploy/message_queues/apache_kafka.py
```diff
@@ -63,20 +63,17 @@ class KafkaMessageQueue(BaseMessageQueue):
         ```
     """
 
-    url: str = DEFAULT_URL
-
     def __init__(
-        self,
-        url: str = DEFAULT_URL,
-        **kwargs: Any,
+        self, config: KafkaMessageQueueConfig = KafkaMessageQueueConfig()
```
logan-markewich: Should an object be a default in kwargs? Doesn't this have the same implications of setting a default to be a dict? (i.e. changes to the default persist throughout runtime)

llama_deploy/apiserver/deployment.py
logan-markewich: if `KafkaMessageQueue` has a single argument in its constructor, do we actually want to model dump here? Shouldn't we just pass the config directly?

masci: Yes, I wanted to keep the changes minimal and I didn't notice but let's fix it

logan-markewich (approved):

PR #355, opened 2024-11-08
docs: Python SDK

masci:
Note: when dependencies change on `llama_deploy` main branch, the lockfile in docs prevents a clean build until you run `poetry lock`. This is not obvious so I decided to remove the lockfile from docs altogether.


docs/docs/module_guides/llama_deploy/40_python_sdk.md
```diff
@@ -1,23 +1,80 @@
...
 ```python
-from llama_deploy import (
-    LlamaDeployClient,
-    AsyncLlamaDeployClient,
-    ControlPlaneConfig,
-)
-
-client = LlamaDeployClient(ControlPlaneConfig())
-async_client = AsyncLlamaDeployClient(ControlPlaneConfig())
+from llama_deploy import Client
```
logan-markewich: since `Client` is such a generic name, I wonder if we should recommend something like 

```
import llama_deploy

client = llama_deploy.Client()
```

or `from llama_deploy import Client as LlamaDeployClient` ?

docs/docs/module_guides/llama_deploy/40_python_sdk.md
logan-markewich, line 1: I feel like this page is missing at least one end-to-end example using the client 🤔 If I landed on this page, I wouldn't really know what this client is for without clicking into the linked API docs and reading every method

docs/docs/module_guides/llama_deploy/40_python_sdk.md
masci: We didn't cut a release yet, I think we're on time to change the class name directly `LlamaDeployClient`

docs/docs/module_guides/llama_deploy/40_python_sdk.md
masci: Added!

docs/docs/module_guides/llama_deploy/40_python_sdk.md
logan-markewich: We can definitely do that :)

logan-markewich (approved):

PR #365, opened 2024-11-13
refact: Use the Python SDK in the CLI implementation

### LINKED ISSUE #336: Use `Client` in the CLI instead of manual http calls
The CLI is basically a frontend to the apiserver, so it could benefit from the `Client` introduced in #290, but currently we make http calls manually, which is prone to error and introduces duplicated code. Let's consolidate using `Client` across the board.

masci:
Fixes #336 

Use the new `Client` class from the CLI


logan-markewich (approved):
Looks great! The CLI code is much easier to understand now

PR #375, opened 2024-11-23
Downgrading python-multipart dependency

tituslhy:
Downgrading python-multipart dependency to avoid a conflict with [Chainlit](https://github.com/Chainlit/chainlit)


masci (changes requested):
Thanks for the PR! Can you run `poetry lock` and check in the new lockfile?

tituslhy:
Done @masci!

logan-markewich (approved):

PR #382, opened 2024-11-26
correct `.yml` file name

ravi03071991:
The `.yml` file was incorrectly referred to as `deployment.yml` instead of `quick_start.yml` in the `README.md` file. This PR updates and corrects the reference.


masci (approved):
Thanks!

PR #386, opened 2024-11-26
fix: fix rabbitmq import errors

### LINKED ISSUE #381: got error ‘Connection’ is not defined llama_deploy rabbitmq
hi, I got error when using llama_deploy[rabbitmq] when run deploy_core. In version 0.2.0 it works but when I try to upgrade to version higher, it got errror because of the TYPE_CHECKING is always False, so it cannot import necessary library. how to solve this problem
IMG_7328

compare to previous version: “return await aio_pika.connect(url))

masci:
fixes #381

Also in this PR:
- added e2e tests
- uniformed the c-tor to the new format we introduced for Kafka

### DIFF
--- a/e2e_tests/message_queues/message_queue_rabbitmq/conftest.py
+++ b/e2e_tests/message_queues/message_queue_rabbitmq/conftest.py
@@ -1,0 +1,22 @@
+import subprocess
+from pathlib import Path
+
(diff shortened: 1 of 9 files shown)


nerdai (approved):
LGTM!

PR #388, opened 2024-11-27
test: fix runtime error in python<3.12

masci:
Clean up loop shutdown in tests, not 100% sure why the problem doesn't affect 3.12
image

### DIFF
--- a/tests/services/test_workflow_service.py
+++ b/tests/services/test_workflow_service.py
@@ -1,15 +1,16 @@
 import asyncio
 import json
-import pytest
-from pydantic import PrivateAttr
 from typing import Any, List
-from llama_index.core.workflow import Workflow, StartEvent, StopEvent, step
-from llama_index.core.workflow.events import HumanResponseEvent, InputRequiredEvent
-from llama_index.core.workflow.context_serializers import JsonSerializer
 
-from llama_deploy.messages import QueueMessage
+import pytest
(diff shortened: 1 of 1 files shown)


tests/services/test_workflow_service.py
```diff
@@ -162,3 +166,6 @@ async def test_hitl_workflow_service(
     result = human_output_consumer.processed_messages[-1]
     assert result.action == ActionTypes.COMPLETED_TASK
     assert result.data["result"] == "42"
+
+    # allow a clean shutdown
+    await asyncio.sleep(0)
```
logan-markewich: I think you still need to await the task after calling cancel right? This lets it shutdown properly

masci: Uhm not sure which task?

logan-markewich: the two tasks that got cancelled

logan-markewich:
```
mq_task.cancel()
server_task.cancel()

await mq_task
await server_task
```

We do something similar in workflows
https://github.com/run-llama/llama_index/blob/55a5c02c434f7db3be13cbedcb7974cf1556b8f2/llama-index-core/llama_index/core/workflow/workflow.py#L428

masci: Ah gotcha! Indeed that's better, PR updated.

logan-markewich (approved):

PR #391, opened 2024-11-27
feat: make the KV store configurable via env vars

### LINKED ISSUE #370: Allow to configure KV store in the control plane via URI
To be able to scale the control plane horizontally, the different instances should be able to share the same metadata storage. This is possible by passing an instance of a `BaseKVStore` to the control plane constructor in the `state_store` parameter, but it's not very practical because you would need some Python glue code to create the instance based on some configuration.

A better approach would be passing an URI describing the KVStore to be used in the control planes, so that each control plane replica could load it from an env var like `CONTROL_PLANE_SERVICE_KV_STORE_URI`.

Feature request coming from #357

masci:
Fixes #370 

Allow passing a URI string to the control plane constructor to build a KV store. For now only Redis and MongoDB are supported, to be considered reference implementation. Ideally we could support all the available store in llama_index.

### DIFF
--- a/examples/redis_state_store/README.md
+++ b/examples/redis_state_store/README.md
@@ -4,7 +4,8 @@
 > This example is mostly based on the [Quick Start](../quick_start/README.md), see there for more details.
 
 We'll be deploying a simple workflow on a local instance of Llama Deploy using Redis as a scalable storage for the
-global state. This is mostly needed when you have more than one control plane running concurrently.
+global state. See [the Control Plane documentation](https://docs.llamaindex.ai/en/stable/module_guides/llama_deploy/20_core_components/#control-plane)
+for an overview of what the global state consists of and when the default storage might not be enough.
 
 Before starting Llama Deploy, use Docker compose to start the Redis container and run it in the background:
 
--- a/examples/redis_state_store/README.md
+++ b/examples/redis_state_store/README.md
@@ -1,0 +1,69 @@
+# Using Redis as State Store
+
+> [!NOTE]
+> This example is mostly based on the [Quick Start](../quick_start/README.md), see there for more details.
+
+We'll be deploying a simple workflow on a local instance of Llama Deploy using Redis as a scalable storage for the
+global state. This is mostly needed when you have more than one control plane running concurrently.
+
+Before starting Llama Deploy, use Docker compose to start the Redis container and run it in the background:
+
+```
+$ docker compose up -d
+```
+
+Make sure to install the package to support the Redis KV store in the virtual environment where we'll run Llama Deploy:
+
+```
+$ pip install -r requirements.txt
+```
+
+This is the code defining our deployment, with comments to the relevant bits:
+
+```yaml
+name: QuickStart
+
+control-plane:
+  port: 8000
+  # Here we tell the Control Plane to use Redis
+  state_store_uri: redis://localhost:6379
+
+default-service: echo_workflow
+
+services:
+  echo_workflow:
+    name: Echo Workflow
+    source:
+      type: local
+      name: ./src
+    path: workflow:echo_workflow
+```
+
+Note how we provide a connection URI for Redis in the `state_store_uri` field of the control plane configuration.
+
+At this point we have all we need to run this deployment. Ideally, we would have the API server already running
+somewhere in the cloud, but to get started let's start an instance locally. Run the following python script
+from a shell:
+
+```
+$ python -m llama_deploy.apiserver
+INFO:     Started server process [10842]
+INFO:     Waiting for application startup.
+INFO:     Application startup complete.
+INFO:     Uvicorn running on http://0.0.0.0:4501 (Press CTRL+C to quit)
+```
+
+From another shell, use the CLI, `llamactl`, to create the deployment:
+
+```
+$ llamactl deploy quick_start.yml
+Deployment successful: QuickStart
+```
+
+Our workflow is now part of the `QuickStart` deployment and ready to serve requests! We can use `llamactl` to interact
+with this deployment:
+
+```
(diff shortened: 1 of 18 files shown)


examples/redis_state_store/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+This is the code defining our deployment, with comments to the relevant bits:
+
+```yaml
+name: QuickStart
+
+control-plane:
+  port: 8000
+  # Here we tell the Control Plane to use Redis
+  state_store_uri: redis://localhost:6379
```
logan-markewich, line 30: so we are going with no explicit label for the state store, and deciding which to use based on params?

examples/redis_state_store/README.md
```diff
@@ -0,0 +1,69 @@
+# Using Redis as State Store
+
+> [!NOTE]
+> This example is mostly based on the [Quick Start](../quick_start/README.md), see there for more details.
+
+We'll be deploying a simple workflow on a local instance of Llama Deploy using Redis as a scalable storage for the
+global state. This is mostly needed when you have more than one control plane running concurrently.
```
logan-markewich: We should add some kind of note that the `workflow` context is what is being maintained between runs/tasks in a single session

examples/redis_state_store/src/workflow.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+# create a dummy workflow
+class EchoWorkflow(Workflow):
+    """A dummy workflow with only one step sending back the input given."""
+
+    @step()
+    async def run_step(self, ev: StartEvent) -> StopEvent:
+        message = str(ev.get("message", ""))
+        return StopEvent(result=f"Message received: {message}")
```
logan-markewich: Maybe a better example would be something that keeps state/context between runs? Like a counter?

llama_deploy/control_plane/config.py
```diff
@@ -21,10 +23,40 @@ class ControlPlaneConfig(BaseSettings):
...
     topic_namespace: str = "llama_deploy"
+    state_store_uri: str | None = None
 
     @property
     def url(self) -> str:
-        if self.port:
-            return f"http://{self.host}:{self.port}"
-        else:
-            return f"http://{self.host}"
+        return f"http://{self.host}:{self.port}"
```
logan-markewich, line 42: Isn't it possible for the port to be blank, if using some actual host name instead of an ip address? I think I added this for a github issue at one point

llama_deploy/control_plane/config.py
```diff
@@ -21,10 +23,40 @@ class ControlPlaneConfig(BaseSettings):
...
+
+
+def parse_state_store_uri(uri: str) -> BaseKVStore:
```
logan-markewich, line 45: This makes sense. Seems easy enough to add others. Although not every kvstore will take only a uri (mongo takes a db_name parameter for example)

llama_deploy/control_plane/config.py
masci: I removed it because the field has a default value, you mean by explicitly set it blank?

llama_deploy/control_plane/config.py
masci: Mongo takes a URI I think, or at least it supports a URI connection string. For databases not supporting a connection string, we can come up with one ourselves, and make the appropriate calls to the underlying KVstore class.

examples/redis_state_store/README.md
masci: My initial idea was to set the class name explicitly here, like `state_store_class: 'llama_index.stores.etc.etc` but then I still needed to collect parameters like host and port, so I went all in with the connection string. The protocol part of the URI should be solid enough to let us pick the right kvstore class.

llama_deploy/control_plane/config.py
logan-markewich: Yea exactly. There was a PR/github issue for this but I can't find it now 😓 But basically, some user had their control plane hosted under `http://my-control-plane` (with no port)

llama_deploy/control_plane/config.py
logan-markewich: fair

llama_deploy/control_plane/config.py
logan-markewich: Not a huge deal either, we can merge this if we think it won't come up again

masci: Actually `port` is an integer so it can't be "blank", the condition would be false only with a port value of 0 or negative. The guy with `http://my-control-plane` must have had it on port 80

examples/redis_state_store/README.md
masci: I went ahead and added a bunch of documentation around this. I'll make this README point to the docs.

examples/redis_state_store/src/workflow.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+# create a dummy workflow
+class CounterWorkflow(Workflow):
+    """A dummy workflow with only one step sending back the input given."""
+
+    @step()
+    async def run_step(self, ctx: Context, ev: StartEvent) -> StopEvent:
+        amount = float(ev.get("amount", 0.0))
+        total = await ctx.get("total", 0.0) + amount
```
logan-markewich, line 13: An await and + without brackets works? TIL :D

logan-markewich (approved):

examples/redis_state_store/src/workflow.py
masci: You made me curious :) indeed `await` is pretty high in the order of precedence, right after call() https://docs.python.org/3/reference/expressions.html#operator-precedence

PR #394, opened 2024-11-29
fix: make SimpleMessageQueue server cancellable

masci:
Several services don't shutdown properly, and while this is not a problem when you tear down the whole process, it's becoming an issue while we start moving towards a more distributed architecture, where services are expected to come and go (think of orchestrators).

This PR makes `SimpleMessageQueue.launch_server()` cancellable, more to come.

### DIFF
--- a/e2e_tests/message_queues/message_queue_simple/test_message_queue.py
+++ b/e2e_tests/message_queues/message_queue_simple/test_message_queue.py
@@ -1,0 +1,19 @@
+import asyncio
+
+import pytest
+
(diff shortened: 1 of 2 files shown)


logan-markewich (approved):
Ah, so basically anything with a lifespan will need this change?

masci:
> Ah, so basically anything with a lifespan will need this change?

It would work anyways, but I'd say, unless the lifespan is needed for also other purposes, keeping the cancelling logic close to the "server launch" logic is easier to maintain

PR #395, opened 2024-11-29
Remove event because of pickle issues

ravi03071991:
PR to fix pickle issues by removing QueryIndexEvent in the example.


masci (approved):
LGTM!

PR #406, opened 2024-12-14
ci: try a different strategy for rabbit health check

masci:
I suspect the healthcheck never really worked because the `CMD-SHELL` was missing from the test command. This would explain why with an interval of 30s it seemed to work (30 seconds are more than enough to boot rabbit and a falsely positive healthcheck would be fine at that point).

### DIFF
--- a/e2e_tests/message_queues/message_queue_rabbitmq/docker-compose.yml
+++ b/e2e_tests/message_queues/message_queue_rabbitmq/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       - "15672:15672"
     healthcheck:
       test:
-        - sh
+        - CMD-SHELL
         - -c
         - |
           rabbitmq-diagnostics -q check_running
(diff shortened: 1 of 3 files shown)


e2e_tests/message_queues/message_queue_rabbitmq/docker-compose.yml
```diff
@@ -6,7 +6,12 @@ services:
       - "5672:5672"
       - "15672:15672"
     healthcheck:
-      test: rabbitmq-diagnostics -q ping
+      test:
```
nerdai, line 9: Aha - nice one!!!

Do we have to change healthcheck command then in our message queue examples for rabbit? I think it uses same command.

nerdai (approved):

e2e_tests/message_queues/message_queue_rabbitmq/docker-compose.yml
masci: Yes, not urgent because the 30 seconds interval is enough to hide the problem, but the current setup is incorrect.

PR #407, opened 2024-12-16
fix: apiserver doesn't set session id when creating a task

masci:
Probably a leftover of the migration to the new Client, not sure why e2e didn't catch it earlier

### DIFF
--- a/e2e_tests/basic_streaming/test_run_client.py
+++ b/e2e_tests/basic_streaming/test_run_client.py
@@ -17,16 +17,11 @@ def test_run_client(services):
     # kick off run
     task_id = session.run_nowait("streaming_workflow", arg1="hello_world")
 
-    num_events = 0
+    progress_received = []
     for event in session.get_task_result_stream(task_id):
         if "progress" in event:
-            num_events += 1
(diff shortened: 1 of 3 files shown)


e2e_tests/basic_streaming/test_run_client.py
```diff
@@ -17,16 +17,11 @@ def test_run_client(services):
...
         if "progress" in event:
-            num_events += 1
-            if num_events == 1:
-                assert event["progress"] == 0.3
-            elif num_events == 2:
-                assert event["progress"] == 0.6
-            elif num_events == 3:
-                assert event["progress"] == 0.9
+            progress_received.append(event["progress"])
+    assert progress_received == [0.3, 0.6, 0.9]
```
masci, line 24: Unrelated to the PR but this test seems to be flaky, I changed the assertion so next time it happens we'll have some more info about which events where received.

nerdai (approved):

nerdai:
I noticed a lot of our e2e tests (in apiserver in particular don't have the e2e purest.mark)

PR #408, opened 2024-12-16
chore: stop using e2e pytest markers

masci:
Some markers were missing from some e2e tests and were not running in the CI.
Since we already segregate e2e tests within the `e2e_tests` folder, there's no actual need for a marker and the risk of forgetting one is pretty high, so let's just kill it and run the tests like `pytest e2e_tests`.


nerdai (commented):
LGTM!

logan-markewich (approved):

PR #409, opened 2024-12-17
fix: reword manual setup docs and use new Client

### LINKED ISSUE #405: Got another bug when run with rabbitmq at version 0.3.6
Hi,

After you resolved the bug the workflow not consume queue and release v0.3.6. I have upgraded and when running I got this bug immediately 

image

the length of list workflow_state_json is 0 so it cannot enter to the condition check is None. So it create loop infinite

masci:
Fixes #405 

- Clarify the manual steps needed to setup an instance of Llama Deploy
- Use the new `Client` class in the examples


logan-markewich (approved):
Nice! One nit, I think we should keep a consistent naming style -- `LlamaIndex`, `LlamaParse` ... `LlamaDeploy`

PR #411, opened 2024-12-18
chore: rearrange simple message queue code

masci:
In preparation of further refactoring, I moved the simple message queue implementation into a dedicated package.
There are no functional changes, and the import paths are backward compatible.


nerdai (commented):
LGTM! ✅

PR #412, opened 2024-12-18
[feat] Add send event endpoints to api server

### LINKED ISSUE #300: Add API Server endpoint to be able to send event to a WorkflowService using `SessionClient.send_event`

nerdai:
This PR adds a new endpoint in api server allowing users to post a new `Event` that gets sent to a running workflow of a specified deployment.

I've also added the `send_event()` method to the `Task` apiserver/model class.

### DIFF
--- a/llama_deploy/apiserver/routers/deployments.py
+++ b/llama_deploy/apiserver/routers/deployments.py
@@ -10,9 +10,10 @@ from llama_deploy.types import (
     DeploymentDefinition,
     SessionDefinition,
     TaskDefinition,
-    EventDefinition,
 )
 from llama_deploy.types.core import TaskResult
+
+from llama_index.core.workflow.events import HumanResponseEvent
 
 deployments_router = APIRouter(
     prefix="/deployments",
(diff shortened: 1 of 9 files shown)


tests/client/models/test_apiserver.py
```diff
@@ -90,7 +90,7 @@ async def test_session_collection_list(client: Any) -> None:
 @pytest.mark.asyncio
 async def test_task_results(client: Any) -> None:
     res = TaskResult(task_id="a_result", history=[], result="some_text", data={})
-    client.request.return_value = mock.MagicMock(json=lambda: res.model_dump_json())
+    client.request.return_value = mock.MagicMock(json=lambda: res.model_dump())
```
nerdai, line 93: I think we had this incorrectly mocked on the basis that `httpx.Response.json()` returns a `dict` rather than a `str`.

I ran into this error when running the new hitl e2e test and the `TaskResult` was failing Pydantic's validation. The modification made in this PR resolves it.

(The naming on Pydantic is confusing... `model_dump` works with `dict` and `model_dump_json` works with `str`.)

llama_deploy/apiserver/routers/deployments.py
```diff
@@ -104,6 +110,26 @@ async def create_deployment_task_nowait(
     return task_definition
 
 
+@deployments_router.post("/{deployment_name}/tasks/{task_id}/human_response_event")
```
nerdai: While the low-level api allows you to send any `Event` type to the workflow service, I think we need to restrict `HumanResponseEvent` unless the api server also contains the module with the custom event?

nerdai: UPDATE:

I was wrong about needing to restrict to `HumanResponseEvent`. The server doesn't need to import the custom Event class, it only needs to forward the event obj str (i.e serialized obj) to the Workflow service.

llama_deploy/client/models/core.py
```diff
@@ -109,6 +109,18 @@ async def send_event(self, service_name: str, task_id: str, ev: Event) -> None:
         url = f"{self.client.control_plane_url}/sessions/{self.id}/tasks/{task_id}/send_event"
         await self.client.request("POST", url, json=event_def.model_dump())
 
+    async def send_event_def(self, task_id: str, ev_def: EventDefinition) -> None:
```
nerdai, line 112: So, I needed to add this in order for the apiserver to be able to just forward along an `EventDefinition` to the WorkflowService

logan-markewich (approved):

masci (approved):
Thanks for the extra comments, LGTM!

PR #414, opened 2024-12-20
test: mock sleep time to .1 seconds to speed up test time

masci:
One of the Solace tests hits a `time.sleep` of 10 seconds, slowing down unit tests. This PR mocks the sleep value to .1 seconds.


logan-markewich (approved):
booo 10s sleeps!

PR #419, opened 2025-01-02
feat: Reload a running deployment.

### LINKED ISSUE #322: Allow re-creating a deployment that already exists
At the moment trying to deploy the same `deployment.yml` results in an error. 

We could leave the default behaviour like it is, but allow users to force a reload of the same deployment  via a `--force` or `--recreate` flag to the CLI that can be replicated in the corresponding apiserver API with a simple query parameter.

masci:
Fixes #322 

Reload a running deployment by restarting its services. The message queue and the control plane won't be restarted, the thread running the dedicated ioloop for this deployment will stay active.

### DIFF
--- a/e2e_tests/apiserver/deployments/deployment1.yml
+++ b/e2e_tests/apiserver/deployments/deployment1.yml
@@ -11,5 +11,5 @@ services:
     host: localhost
     source:
       type: git
-      name: https://github.com/run-llama/llama_deploy.git
-    path: tests/apiserver/data/workflow:my_workflow
+      name: https://github.com/run-llama/llama_deploy.git@massi/322
+    path: e2e_tests/apiserver/deployments/src:my_workflow
--- a/llama_deploy/apiserver/deployment.py
+++ b/llama_deploy/apiserver/deployment.py
@@ -124,6 +124,9 @@ class Deployment:
         # Run allthethings
         await asyncio.gather(*tasks)
 
+    def reload(self) -> None:
+        pass
+
     def _load_services(self, config: Config) -> list[WorkflowService]:
         """Creates WorkflowService instances according to the configuration object."""
         workflow_services = []
@@ -278,7 +281,7 @@ class Manager:
         self._deployments_path = deployments_path
         self._max_deployments = max_deployments
         self._pool = ThreadPool(processes=max_deployments)
-        self._control_plane_port = 8002
+        self._last_control_plane_port = 8002
 
     @property
     def deployment_names(self) -> list[str]:
@@ -297,34 +300,46 @@ class Manager:
         except asyncio.CancelledError:
             pass
 
-    def deploy(self, config: Config) -> None:
+    def deploy(self, config: Config, reload: bool = False) -> None:
         """Creates a Deployment instance and starts the relative runtime.
 
         Args:
             config: The deployment configuration.
+            reload: Reload an existing deployment instead of raising an error.
 
         Raises:
             ValueError: If a deployment with the same name already exists or the maximum number of deployment exceeded.
             DeploymentError: If it wasn't possible to create a deployment.
         """
-        if config.name in self._deployments:
-            msg = f"Deployment already exists: {config.name}"
-            raise ValueError(msg)
+        if not reload:
+            # Raise an error if deployment already exists
+            if config.name in self._deployments:
+                msg = f"Deployment already exists: {config.name}"
+                raise ValueError(msg)
 
-        if len(self._deployments) == self._max_deployments:
-            msg = "Reached the maximum number of deployments, cannot schedule more"
-            raise ValueError(msg)
+            # Raise an error if we can't create any new deployment
+            if len(self._deployments) == self._max_deployments:
+                msg = "Reached the maximum number of deployments, cannot schedule more"
+                raise ValueError(msg)
 
-        self._assign_control_plane_address(config)
+            # Set the control plane TCP port in the config where not specified
+            self._assign_control_plane_address(config)
 
-        deployment = Deployment(config=config, root_path=self._deployments_path)
-        self._deployments[config.name] = deployment
-        self._pool.apply_async(func=asyncio.run, args=(deployment.start(),))
+            deployment = Deployment(config=config, root_path=self._deployments_path)
+            self._deployments[config.name] = deployment
+            self._pool.apply_async(func=asyncio.run, args=(deployment.start(),))
+        else:
+            if config.name not in self._deployments:
+                msg = f"Cannot find deployment to reload: {config.name}"
+                raise ValueError(msg)
+
+            deployment = self._deployments[config.name]
+            deployment.reload()
 
     def _assign_control_plane_address(self, config: Config) -> None:
         for service in config.services.values():
             if not service.port:
-                service.port = self._control_plane_port
-                self._control_plane_port += 1
+                service.port = self._last_control_plane_port
+                self._last_control_plane_port += 1
             if not service.host:
                 service.host = "localhost"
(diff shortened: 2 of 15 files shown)


e2e_tests/apiserver/deployments/deployment1.yml
```diff
@@ -11,5 +11,5 @@ services:
     host: localhost
     source:
       type: git
-      name: https://github.com/run-llama/llama_deploy.git
-    path: tests/apiserver/data/workflow:my_workflow
+      name: https://github.com/run-llama/llama_deploy.git@massi/322
+    path: e2e_tests/apiserver/deployments/src:my_workflow
```
masci, line 15: There is a catch-22 in here, the tests use this repo but until we merge this PR on `main` I need to use my branch. When we'll merge this PR, Github will delete the branch and tests will need an update 😅

llama_deploy/apiserver/deployment.py
```diff
@@ -112,17 +117,35 @@ async def start(self) -> None:
...
+            for wfs in self._workflow_services:
+                service_task = asyncio.create_task(wfs.launch_server())
+                self._service_tasks.append(service_task)
+                consumer_fn = await wfs.register_to_message_queue()
+                control_plane_url = f"http://{self._control_plane_config.host}:{self._control_plane_config.port}"
+                await wfs.register_to_control_plane(control_plane_url)
+                consumer_task = asyncio.create_task(consumer_fn())
+                self._service_tasks.append(consumer_task)
+            self._service_startup_complete.set()
+            await asyncio.gather(*self._service_tasks)
```
masci, line 164: This is the key change. We assume that until `self._running` is true, if the services tasks are being cancelled they'll be restarted. We send the signal `_service_startup_complete` when the load process is done so that the `reload` function will know it can return.

llama_deploy/apiserver/deployment.py
```diff
@@ -112,17 +117,35 @@ async def start(self) -> None:
...
-            await wfs.register_to_control_plane(control_plane_url)
-            consumer_task = asyncio.create_task(consumer_fn())
-            tasks.append(consumer_task)
+        tasks.append(asyncio.create_task(self._run_services()))
 
         # Run allthethings
         await asyncio.gather(*tasks)
+        self._running = False
+
+    async def reload(self, config: Config) -> None:
```
logan-markewich, line 126: I'm actually slightly confused, but how does this reload things? I see only cancelling and then waiting for a flag? (Just trying to follow the reload logic)

llama_deploy/apiserver/deployment.py
```diff
@@ -112,17 +117,35 @@ async def start(self) -> None:
...
+        """Reload this deployment by restarting its services."""
+        self._workflow_services = self._load_services(config)
+        self._default_service = config.default_service
+
+        for t in self._service_tasks:
+            t.cancel()
```
logan-markewich, line 139: Should we also `await` each task after it cancels so that it shutdowns properly?

llama_deploy/apiserver/deployment.py
```diff
@@ -151,9 +174,13 @@ def _load_services(self, config: Config) -> list[WorkflowService]:
                 raise ValueError(msg)
 
             # Sync the service source
-            destination = self._path / service_id
+            destination = (self._path / service_id).resolve()
+
+            if destination.exists():
+                rmtree(str(destination))
```
logan-markewich, line 202: What are we writing to disk? 👀 (I may have missed when this was introduced)

masci, line 202: Not sure if this should be somehow configurable, or if we should delegate this operation to the source managers, but for now this works - at every deploy, any data used by the source manager in the file system is wiped, so the next reload won't fail

llama_deploy/apiserver/deployment.py
```diff
@@ -297,34 +324,46 @@ async def serve(self) -> None:
...
+            reload: Reload an existing deployment instead of raising an error.
 
         Raises:
             ValueError: If a deployment with the same name already exists or the maximum number of deployment exceeded.
             DeploymentError: If it wasn't possible to create a deployment.
         """
-        if config.name in self._deployments:
-            msg = f"Deployment already exists: {config.name}"
-            raise ValueError(msg)
+        if not reload:
```
masci, line 360: This if branch behaves exactly as before

llama_deploy/apiserver/deployment.py
```diff
@@ -297,34 +324,46 @@ async def serve(self) -> None:
...
+                raise ValueError(msg)
 
-        self._assign_control_plane_address(config)
+            # Set the control plane TCP port in the config where not specified
+            self._assign_control_plane_address(config)
+
+            deployment = Deployment(config=config, root_path=self._deployments_path)
+            self._deployments[config.name] = deployment
+            self._pool.apply_async(func=asyncio.run, args=(deployment.start(),))
+        else:
```
masci, line 377: this branch is new

llama_deploy/apiserver/deployment.py
```diff
@@ -297,34 +324,46 @@ async def serve(self) -> None:
...
+            deployment = self._deployments[config.name]
+            await deployment.reload(config)
 
     def _assign_control_plane_address(self, config: Config) -> None:
         for service in config.services.values():
             if not service.port:
-                service.port = self._control_plane_port
-                self._control_plane_port += 1
+                service.port = self._last_control_plane_port
+                self._last_control_plane_port += 1
```
masci, line 389: unrelated, I changed the name of the variable for clarity

llama_deploy/apiserver/deployment.py
```diff
@@ -112,17 +117,35 @@ async def start(self) -> None:
...
+
+    async def reload(self, config: Config) -> None:
+        """Reload this deployment by restarting its services."""
+        self._workflow_services = self._load_services(config)
+        self._default_service = config.default_service
+
+        for t in self._service_tasks:
+            t.cancel()
+
+        await self._service_startup_complete.wait()
```
masci, line 142: The logic here is to cancel the running services and wait for the reload in `_run_services` to happen.

llama_deploy/apiserver/deployment.py
masci: The tasks are already awaited by `_run_services`, so in this function we can just skip that and await for the final signal

llama_deploy/apiserver/deployment.py
masci: Yep, we cancel all the service tasks and wait for the while loop in `_run_services` to restart them

llama_deploy/apiserver/deployment.py
masci: In the case of the git source manager, it's the git repo, for the local manager it's the folder specified in the deployment file

llama_deploy/apiserver/deployment.py
logan-markewich: ah ok, gotcha! -- can we maybe add a comment for that?

llama_deploy/apiserver/deployment.py
logan-markewich: ah gotcha. So redeploying means the repo is cloned again or the file is copied back in 👍🏻

llama_deploy/apiserver/deployment.py
masci: Done!

llama_deploy/apiserver/deployment.py
masci: added a comment

logan-markewich (approved):

PR #422, opened 2025-01-07
fix: Invoke cleanup() method on the message queue from deploy() before exiting

masci:
Message queue clients have a method that's supposed to be invoked on exit to clean up the queue if needed (depending on the queue implementation). The method wasn't called from `deploy()` and while this is ok-ish most of the times, while working on the Redis message queue I noted the redis client is not happy without a proper shutdown of its client connections, and my e2e tests were failing.

With this PR:
- `cleanup_local` was renamed to `cleanup` to avoid confusion (cleanup also happens when server is remote)
- `cleanup` is called from `deploy()` on shutdown

This way, all the e2e tests currently relying on `deploy()` will implicitly clean up the queue on exit, improving coverage.

Also in this PR:
- The shutdown procedure in `deploy()` was refactored to rely on `CancelledError` instead of `SIGTERM` in a more asyncio-friendly fashion
- The control plane server now exits gracefully when interrupted
- The e2e tests modules for the message queues where renamed to avoid stuttering


e2e_tests/message_queues/kafka/__init__.py
logan-markewich, line 1: This means we no longer have "examples" of other message queues? We might want to add this to the docs at some point, I've had a few people message me asking how to configure the message queue in the yaml

(the answer I give is using the `message_queue:` key and then any settings from their config class in the api reference)

llama_deploy/message_queues/base.py
```diff
@@ -96,11 +91,3 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
     @abstractmethod
     async def processing_loop(self) -> None:
         """The processing loop for the service."""
-
-    @abstractmethod
-    async def launch_local(self) -> asyncio.Task:
```
logan-markewich, line 101: Not needed anymore since refactoring the simple message queue/client/server ?

logan-markewich (approved):
LGTM!

llama_deploy/message_queues/base.py
masci: yeah, we could start the simple message queue server from there but it seems overkill having it as an abstract method in the base class

PR #427, opened 2025-01-13
feat: support HTTPS control plane URL

### LINKED ISSUE #420: Support HTTPs for Agent -> Control Plane connection
There are several places in the code which hard code "http" when rendering the URL to communicate with the control plane. We are running the control plane behind a k8s ingress controller which exposes a https endpoint and so need the agents to be configured to use https.

We are currently carrying a couple of fixes to make this work:

1) we have a sub-class of the ControlPlaneConfig object which introduces a new option "USE_TLS":

```
class ControlPlaneConfig(llama_deploy.ControlPlaneConfig):
    use_tls: bool = False

    @property
    def url(self) -> str:
        if self.use_tls:
            return f"https://{self.host}:{self.port}"
        return f"http://{self.host}:{self.port}"

```

2) we're carryng a copy of the "deploy_workflow" function which removes these lines because the url is already fetched from the config higher up in the function https://github.com/run-llama/llama_deploy/blob/4c5b832cc262dd35be28086ca6fba0d52d9f8ee5/llama_deploy/deploy/deploy.py#L227-L229

Tehsmash:
This change introduces a new configuration option "USE_TLS" for the ControlPlaneConfig. When True this will render the control plane URL with the HTTPS scheme instead of HTTP.

Fixes: #420

### DIFF
--- a/llama_deploy/control_plane/config.py
+++ b/llama_deploy/control_plane/config.py
@@ -36,9 +36,15 @@ class ControlPlaneConfig(BaseSettings):
         default=None,
         description="The connection URI of the database where to store state. If None, SimpleKVStore will be used",
     )
+    use_tls: bool = Field(
+        default=False,
+        description="Use TLS (HTTPS) to communicate with the control plane",
+    )
 
     @property
     def url(self) -> str:
+        if self.use_tls:
+            return f"https://{self.host}:{self.port}"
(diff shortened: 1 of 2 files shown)


masci (approved):
LGTM!

masci:
Tests failures are on me 😅 merging as soon as it gets green!

PR #441, opened 2025-01-29
fix: avoid 409 errors on message queue when services restart

masci:
This PR addresses several issues:
- creating a topic that already exists in SimpleMessageQueue now is a no-op instead of returning 409
- `deploy_workflow` now cleans up asyncio tasks at shutdown
- the SimpleMessageQueue client is now more resilient at shutdown, catching multiple exceptions
- there was a bug in the payload sent for deregistering a consumer from the control plane
- a method for deregistering consumers from message queues was added
- the Workflow service now properly shuts down


llama_deploy/message_queues/simple/client.py
```diff
@@ -61,17 +61,32 @@ async def start_consuming_callable() -> None:
...
+                        message = QueueMessage.model_validate(result.json())
+                        await consumer.process_message(message)
+                    await asyncio.sleep(0.1)
+
+                except httpx.HTTPError as e:
+                    logger.error(f"HTTP error occurred: {e}")
+                    await asyncio.sleep(1)  # Back off on errors
+                    continue
+
+                except asyncio.CancelledError:
```
logan-markewich, line 79: Nice!

logan-markewich (approved):
A lot of good fixes in here!

PR #443, opened 2025-01-29
feat: create deployments at API Server startup

masci:
Main feature: allow the API Server to run deployments found in a certain folder (defaults to `.llama_deploy_rc` on disk. This is particularly useful to encapsulate a LlamaDeploy instance in a Docker image.

Also in this PR:
- Renamed `Config` to `DeploymentConfig` for clarity
- While creating a deployment, wait for the control plane and the workflow services to be responsive before returning
- Adapted the end-to-end tests to use all the above
- Added some additional unit tests

### DIFF
--- a/llama_deploy/apiserver/deployment_config_parser.py
+++ b/llama_deploy/apiserver/deployment_config_parser.py
@@ -64,7 +64,7 @@ class Service(BaseModel):
     ts_dependencies: dict[str, str] | None = Field(None, alias="ts-dependencies")
 
 
-class Config(BaseModel):
+class DeploymentConfig(BaseModel):
     """Model definition mapping a deployment config file."""
 
     name: str
--- a/llama_deploy/apiserver/deployment_config_parser.py
+++ b/llama_deploy/apiserver/deployment_config_parser.py
@@ -1,11 +1,11 @@
+import sys
 from enum import Enum
 from pathlib import Path
 from typing import Annotated, Union
-import sys
 
 if sys.version_info >= (3, 11):
     from typing import Self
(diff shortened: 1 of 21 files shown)


llama_deploy/apiserver/deployment_config_parser.py
```diff
@@ -64,7 +63,7 @@ class Service(BaseModel):
     ts_dependencies: dict[str, str] | None = Field(None, alias="ts-dependencies")
 
 
-class Config(BaseModel):
+class DeploymentConfig(BaseModel):
```
logan-markewich, line 66: ooc, why the rename?

logan-markewich: I guess to better differentiate bewtween config, settings, etc. ?

llama_deploy/apiserver/server.py
```diff
@@ -19,8 +20,25 @@
...
+    settings = ApiserverSettings()
     t = manager.serve()
+    logger.info(f"rc folder: {settings.rc_path}")
+
+    if settings.rc_path.exists():
+        logger.info(
+            f"Browsing the rc folder {settings.rc_path} for deployments to start"
+        )
+        # match both .yml and .yaml files with the glob
+        for yaml_file in Path(settings.rc_path).glob("*.y*ml"):
```
logan-markewich: isn't settings.rc_path already a Path object?

logan-markewich (approved):
pretty neat!

hristogg:
Thanks! WIll test it out!

PR #449, opened 2025-02-04
chore: refactor WorkflowService configuration

masci:
Enforce a single source of truth for the `WorkflowService` configuration through the `WorkflowServiceConfig` object. Base class doesn't need to be Pydantic model anymore.

### DIFF
--- a/llama_deploy/services/base.py
+++ b/llama_deploy/services/base.py
@@ -3,7 +3,6 @@ from abc import ABC, abstractmethod
 from typing import Any
 
 import httpx
-from pydantic import BaseModel, ConfigDict, PrivateAttr
 
 from llama_deploy.control_plane.server import ControlPlaneConfig
 from llama_deploy.message_consumers.base import (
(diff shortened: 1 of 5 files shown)


llama_deploy/services/base.py
```diff
@@ -31,15 +30,19 @@ class BaseService(MessageQueuePublisherMixin, ABC, BaseModel):
     - A service can be registered to the message queue.
     """
 
-    def __init__(self, *args: Any, **kwargs: Any) -> None:
-        super().__init__(*args, **kwargs)
+    def __init__(
+        self,
+        name: str,
+        control_plane_url: str | None = None,  # deprecated
+        control_plane_config: ControlPlaneConfig = ControlPlaneConfig(),
```
logan-markewich: should this be `= None` instead?

logan-markewich: the later on we do `control_plane_config or ControlPlaneConfig()`

PR #450, opened 2025-02-06
refact: support absolute paths in local source manager

masci:
This PR introduced the concept that the path for a service with a local source is relative to the deployment file. For example, if you have a deployment file `/foo/bar/deployment.yml` containing the following:
```yaml
services:
  dummy_workflow:
    name: Dummy Workflow
    source:
      type: local
      name: src  # Relative to what?
    path: workflow:echo_workflow
```
the local source manager will look for `src` in `/foo/bar/src`. 

This change required a refactoring of the source managers: 
- managers now take a `DeploymentConfig` instance in the constructor, hence the protocol was changed into an abstract class
- the deployment config has a new field `base_path` that's automatically populated when `from_yaml` is used.

### DIFF
--- a/llama_deploy/apiserver/deployment_config_parser.py
+++ b/llama_deploy/apiserver/deployment_config_parser.py
@@ -71,6 +71,7 @@ class DeploymentConfig(BaseModel):
     message_queue: MessageQueueConfig | None = Field(None, alias="message-queue")
     default_service: str | None = Field(None, alias="default-service")
     services: dict[str, Service]
+    base_path: Path = Path()
 
     @classmethod
     def from_yaml_bytes(cls, src: bytes) -> Self:
@@ -83,4 +84,4 @@ class DeploymentConfig(BaseModel):
         """Read config data from a yaml file."""
         with open(path, "r") as yaml_file:
             config = yaml.safe_load(yaml_file) or {}
-        return cls(**config)
+        return cls(**config, base_path=path.parent)
--- a/e2e_tests/apiserver/rc/deployment.yml
+++ b/e2e_tests/apiserver/rc/deployment.yml
@@ -11,5 +11,5 @@ services:
     host: localhost
     source:
       type: local
-      name: ./e2e_tests/apiserver/rc/src
+      name: src
     path: workflow:echo_workflow
--- a/llama_deploy/apiserver/deployment.py
+++ b/llama_deploy/apiserver/deployment.py
@@ -6,7 +6,7 @@ import sys
 from multiprocessing.pool import ThreadPool
 from pathlib import Path
 from shutil import rmtree
-from typing import Any
+from typing import Any, Type
 
 import httpx
 from dotenv import dotenv_values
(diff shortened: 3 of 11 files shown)


llama_deploy/apiserver/deployment_config_parser.py
```diff
@@ -71,6 +71,7 @@ class DeploymentConfig(BaseModel):
     message_queue: MessageQueueConfig | None = Field(None, alias="message-queue")
     default_service: str | None = Field(None, alias="default-service")
     services: dict[str, Service]
+    base_path: Path = Path()
```
logan-markewich, line 74: What is an empty `path` ? Just `""` ?

llama_deploy/apiserver/deployment_config_parser.py
```diff
@@ -83,4 +84,4 @@ def from_yaml(cls, path: Path) -> Self:
         """Read config data from a yaml file."""
         with open(path, "r") as yaml_file:
             config = yaml.safe_load(yaml_file) or {}
-        return cls(**config)
+        return cls(**config, base_path=path.parent)
```
logan-markewich, line 87: If the user sets an absolute path in their yml, will that break because we fore relative here?

llama_deploy/apiserver/deployment_config_parser.py
logan-markewich: I guess this doesn't really matter lol

llama_deploy/apiserver/deployment_config_parser.py
logan-markewich: I guess they would need to set `base_path: /` ?

logan-markewich (approved):
Seems good to me, sensible change 👍🏻 I agree by default it should be relative to the yml

Main worry is that maybe we can auto-detect relative vs. absolute paths

llama_deploy/apiserver/deployment_config_parser.py
masci: It's equivalent to `.`, which was hardcoded before this PR

llama_deploy/apiserver/deployment_config_parser.py
masci: good point, I think we should support both. Do you think users will understand that `/foo` is absolute and `foo` is relative to the deployment file?

logan-markewich: I think they should, seems like a basic feature of how paths work 🤔 `foo` and `./foo` are relative, `/foo` is absolute

masci: Both relative and absolute paths work and I added unit tests to cover that.

So to recap:
- `/path/to/sources` points to the absolute location
- `path/to/sources` refers to the folder containing the deployment file
- `./path/to/sources` same as above

masci:
Pasting here for visibility:

So to recap:

 - `/path/to/sources` points to the absolute location
 - `path/to/sources` refers to the folder containing the deployment file
 - `./path/to/sources` same as above

PR #452, opened 2025-02-07
Pass client timeout to underlying client

apost71:
This is a PR to address https://github.com/run-llama/llama_deploy/issues/374

@masci I think you're right about the fine grained timeout configurations.  Adding that capability introduces difficulties with serialization and ease of use via environment variables and may not be worth the effort.

### DIFF
--- a/llama_deploy/client/async_client.py
+++ b/llama_deploy/client/async_client.py
@@ -134,7 +134,7 @@ class AsyncSessionClient:
         start_time = time.time()
         while True:
             try:
-                async with httpx.AsyncClient() as client:
+                async with httpx.AsyncClient(timeout=self.timeout) as client:
                     async with client.stream(
                         "GET",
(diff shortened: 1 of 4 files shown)


masci (changes requested):
Thanks for the PR! Would you mind also adding the timeout to the async client here https://github.com/run-llama/llama_deploy/blob/main/llama_deploy/client/async_client.py#L137? We should really deprecate those two classes in favor of Client but until then it's maybe better to stay consistent

apost71:
> Thanks for the PR! Would you mind also adding the timeout to the async client here https://github.com/run-llama/llama_deploy/blob/main/llama_deploy/client/async_client.py#L137? We should really deprecate those two classes in favor of Client but until then it's maybe better to stay consistent

Good catch, updated.

masci (approved):
Thank you, merging!

PR #454, opened 2025-02-10
Add Solace Message Queue To API Reference Documents

alimosaed:
**What is the purpose of this change?**
Add Solace message queue to the API reference documents.

**How is this accomplished?**
Added the Solace message queue path to the Solace.md file to automatically generate documents.

**Anything reviews should focus on/be aware of?**
Nothing


masci (approved):
LGTM!

PR #460, opened 2025-02-13
Allow all supported Redis URLs for Control Plane State Store URI.

### LINKED ISSUE #459: Unable to Use TLS / SSL for the Control Plane Redis URI
We encountered an error when setting the `CONTROL_PLANE_STATE_STORE_URI` environment  variable to a Redis cache with TLS / SSL:

```
ValueError: key-value store 'rediss' is not supported.
```

I believe the issue is in parse_state_store_uri function of llama_deploy/control_plane/config.py.

I'm going to do a quick PR to address this.

kdbrooks:
Fixes #459 

Allow all supported Redis URLs for Control Plane State Store URI.

### DIFF
--- a/llama_deploy/control_plane/config.py
+++ b/llama_deploy/control_plane/config.py
@@ -63,14 +63,16 @@ class ControlPlaneConfig(BaseSettings):
 def parse_state_store_uri(uri: str) -> BaseKVStore:
     bits = urlparse(uri)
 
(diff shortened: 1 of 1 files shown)


masci (approved):
Thank you!

Coveralls is currently under an outage, jobs are failing but tests are green, merging...

PR #461, opened 2025-02-13
fix: compute URLs returned in root path

masci:
The root path of the apiserver `/` returns the link to the status endpoint and the link to the OpenAPI docs endpoint. Currently they are hardcoded, meaning they don't work when user changes the host or port of the API Server. This PR computes the URL dynamically.

### DIFF
--- a/llama_deploy/apiserver/app.py
+++ b/llama_deploy/apiserver/app.py
@@ -3,6 +3,7 @@ import os
 
 from fastapi import FastAPI
 from fastapi.middleware.cors import CORSMiddleware
+from fastapi.requests import Request
 from fastapi.responses import JSONResponse
 
 from .routers import deployments_router, status_router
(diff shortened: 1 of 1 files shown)


llama_deploy/apiserver/app.py
```diff
@@ -28,10 +29,10 @@
...
 
 @app.get("/")
-async def root() -> JSONResponse:
+async def root(request: Request) -> JSONResponse:
     return JSONResponse(
         {
-            "swagger_docs": "http://localhost:4501/docs",
-            "status": "http://localhost:4501/status",
+            "swagger_docs": f"{request.base_url}docs",
+            "status": f"{request.base_url}status",
```
logan-markewich, line 36: If the API server is deployed behind a proxy or similar, will this still have the correct URL?

logan-markewich (approved):

llama_deploy/apiserver/app.py
masci: Yes, as long as `X-Forwarded-Proto` and `X-Forwarded-Host` headers are correct (which should be the case if you're using a proxy server)

PR #463, opened 2025-02-13
feat: track deployment and service status with Prometheus

masci:
After requesting the API Server to create a deployment, some time can pass in order to fetch the code, install dependencies and setting up all the LlamaDeploy components that are needed. 

With this PR we track in Prometheus the intermediate states for the API Server itself, a deployment and its services.
We also need better logging but I won't add it here to ease the review.

Resulting metrics:
```
# HELP deployment_state Current state of a deployment
# TYPE deployment_state gauge
deployment_state{deployment_name="QuickStart",deployment_state="loading_services"} 0.0
deployment_state{deployment_name="QuickStart",deployment_state="ready"} 0.0
deployment_state{deployment_name="QuickStart",deployment_state="starting_services"} 1.0
deployment_state{deployment_name="QuickStart",deployment_state="running"} 0.0
deployment_state{deployment_name="QuickStart",deployment_state="stopped"} 0.0
# HELP service_state Current state of a service attached to a deployment
# TYPE service_state gauge
service_state{deployment_name="QuickStart",service_name="dummy_workflow",service_state="loading"} 0.0
service_state{deployment_name="QuickStart",service_name="dummy_workflow",service_state="syncing"} 0.0
service_state{deployment_name="QuickStart",service_name="dummy_workflow",service_state="installing"} 0.0
service_state{deployment_name="QuickStart",service_name="dummy_workflow",service_state="ready"} 1.0
```
(when a state goes to 1.0 the others switch to 0.0)

### DIFF
--- a/llama_deploy/apiserver/deployment.py
+++ b/llama_deploy/apiserver/deployment.py
@@ -40,6 +40,7 @@ from .deployment_config_parser import (
     SourceType,
 )
 from .source_managers import GitSourceManager, LocalSourceManager, SourceManager
+from .stats import deployment_state, service_state
 
 logger = logging.getLogger()
 SOURCE_MANAGERS: dict[SourceType, Type[SourceManager]] = {
@@ -83,6 +84,7 @@ class Deployment:
         self._workflow_services: dict[str, WorkflowService] = self._load_services(
             config
         )
+        deployment_state.labels(self._name).state("ready")
 
     @property
     def default_service(self) -> str | None:
@@ -125,7 +127,9 @@ class Deployment:
             tasks.append(asyncio.create_task(self._run_services()))
 
         # Run allthethings
+        deployment_state.labels(self._name).state("running")
         await asyncio.gather(*tasks)
+        deployment_state.labels(self._name).state("stopped")
         self._running = False
 
     async def reload(self, config: DeploymentConfig) -> None:
@@ -173,6 +177,7 @@ class Deployment:
         if they are all cancelled. This is to support the reload process
         (see reload() for more details).
         """
+        deployment_state.labels(self._name).state("starting_services")
         while self._running:
             self._service_tasks = []
             # If this is a reload, self._workflow_services contains the updated configurations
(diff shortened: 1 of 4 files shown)


llama_deploy/apiserver/deployment.py
```diff
@@ -125,7 +127,9 @@ async def start(self) -> None:
             tasks.append(asyncio.create_task(self._run_services()))
 
         # Run allthethings
+        deployment_state.labels(self._name).state("running")
```
logan-markewich, line 130: Do we need this here if `_run_services` is also setting the state as it runs?

From my understanding, the flow is

running -> starting_services ->loading_services -> loading -> syncing -> installing -> ready -> stopped

llama_deploy/apiserver/deployment.py
```diff
@@ -204,8 +209,10 @@ async def _run_services(self) -> None:
 
     def _load_services(self, config: DeploymentConfig) -> dict[str, WorkflowService]:
         """Creates WorkflowService instances according to the configuration object."""
+        deployment_state.labels(self._name).state("loading_services")
         workflow_services = {}
         for service_id, service_config in config.services.items():
+            service_state.labels(self._name, service_id).state("loading")
```
logan-markewich, line 215: Should this log be specific to the current service? (since these states are all happening under a loop?)

llama_deploy/apiserver/deployment.py
masci: The idea is you can query the three statuses separately:
- Apiserver: starting -> running -> stopped
- Deployment: loading_services -> ready -> starting_services -> running -> stopped
- Service: loading -> syncing -> installing -> ready

For a healthy deployment you would have:
- Apiserver starting -> Deployment loading_services -> (Service statuses for each service) -> Deployment ready -> Deployment starting services -> Deployment running

llama_deploy/apiserver/deployment.py
masci: Uhm the status is labelled after (deployment, service) so it would be specific for current service

logan-markewich: ah right, missed the service_id in the label lol

llama_deploy/apiserver/deployment.py
logan-markewich: Ah I see, ok I think this makes sense

logan-markewich (approved):

Issue #464, opened 2025-02-14
got error related to pydantic at method get_workflow_state


volethanh:
Hi,

I’ve got this issue from the previous version until now (0.6.1)

It’s related to pydantic WorkflowState.model_validate_json whenever the message comes to workflow

![Image](https://github.com/user-attachments/assets/1ad60791-2372-4193-8a69-762e0ea1c311)

I have modify the code to workaround by adding more condition to check 
`len(workflow_state_json) == 0`

![Image](https://github.com/user-attachments/assets/4d113b44-c47d-4a26-84fc-f98d7924fe95)

I don’t know if anyone here got the same error like me ? Then should it be added to the codebase of this package?

logan-markewich:
Hmm that's weird, why was the state a dict instead of a string 🤔

volethanh:
> Hmm that's weird, why was the state a dict instead of a string 🤔

I have defined deploy_core and deploy_workflow as usual 

Then, I use postman to call that workflow

First request is to create sessionId
![Image](https://github.com/user-attachments/assets/aecac402-bbcd-4570-b5e0-3c649f4ffbc0)

Then second request is to call and pass param to the specific workflow

![Image](https://github.com/user-attachments/assets/c4eb0106-7324-4b23-a2ca-04da2ffdf930)

masci:
@volethanh I think the problem is you're passing `sessionID` in the `task_id` field in your second request. I opened https://github.com/run-llama/llama_deploy/pull/465 to raise a specific error when this happens.

masci:
I'm closing this, feel free to reopen if the problem persists

PR #465, opened 2025-02-14
fix: raise an error when task.session_id doesn't match session_id

masci:
Related to #464 

- Make `add_task_to_session` return an error when the request is invalid
- Make `process_message` more robust
- Add a bunch of unit tests, some were low-hanging fruits unrelated to this PR

### DIFF
--- a/llama_deploy/control_plane/server.py
+++ b/llama_deploy/control_plane/server.py
@@ -229,17 +229,19 @@ class ControlPlaneServer(BaseControlPlane):
         return self._publish_callback
 
     async def process_message(self, message: QueueMessage) -> None:
-        action = message.action
+        if not message.data:
+            raise ValueError(f"Invalid field 'data' in QueueMessage: {message.data}")
 
-        if action == ActionTypes.NEW_TASK and message.data is not None:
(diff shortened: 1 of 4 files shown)


logan-markewich (approved):
This makes sense. Still slightly concerned how the original user got a dict when it should be a json string, but this change also makes sense 💪

masci:
@logan-markewich I should've probably added it to the PR description.
Without this check https://github.com/run-llama/llama_deploy/pull/465/files#diff-b0b865e8343afe7478a9166d6ba19e4f382d8e822f9c6c0d56843a6b8e5ef65cR388 we propagate the non-existent task id to the control plane and eventually store in the kv-store.
Then you get here https://github.com/run-llama/llama_deploy/blob/4016fe3fadb375faf244f39d566c99d2fffcf508/llama_deploy/orchestrators/simple.py#L47 where the state is assigned to an empty dict.
I wasn't confident enought to change `state[task_def.task_id] = {}` so I preferred to trap the error earlier, but probably we should have a look at that part of the code to see if we need that fallback assignment in the first place.

PR #469, opened 2025-02-17
feat: proxy Prometheus metrics endpoint through API Server

masci:
In certain serverless environments like Knative and Cloud Run, it's not possible to expose more than one port from the LlamaDeploy container. In case users still want to scrape metrics, we proxy the Prometheus metrics endpoint through the `/status/metrics` endpoint of the API Server.

### DIFF
--- a/llama_deploy/apiserver/routers/status.py
+++ b/llama_deploy/apiserver/routers/status.py
@@ -36,7 +36,7 @@ async def metrics() -> PlainTextResponse:
 
     try:
         async with httpx.AsyncClient() as client:
-            response = await client.get(f"http://localhost:{settings.prometheus_port}/")
+            response = await client.get(f"http://127.0.0.1:{settings.prometheus_port}/")
             return PlainTextResponse(content=response.text)
     except httpx.RequestError as exc:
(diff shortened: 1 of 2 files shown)


llama_deploy/apiserver/routers/status.py
```diff
@@ -16,3 +20,23 @@ async def status() -> Status:
...
+    container cannot expose more than one port (e.g. Knative, Google Cloud Run).
+    If Prometheus is not enabled, this endpoint returns an empty HTTP-204 response.
+    """
+    settings = ApiserverSettings()
+    if not settings.prometheus_enabled:
+        return PlainTextResponse(status_code=204)
+
+    try:
+        async with httpx.AsyncClient() as client:
+            response = await client.get(f"http://localhost:{settings.prometheus_port}/")
```
logan-markewich: just confirming, but hardcoding to localhost is ok to do?

logan-markewich (approved):

llama_deploy/apiserver/routers/status.py
masci: Ah it's not actually, we need 127.0.0.1 there, great catch!

PR #480, opened 2025-03-07
fix: create path to config dir if does not exist

masci:
The config path returned by `platformdirs` might not exist (the most likely scenario in docker containers), make sure we create the full path to the default config file when the CLI runs for the first time.


logan-markewich (approved):
LGTM

PR #485, opened 2025-03-21
Add support for exclusive consumer mode for Redis Message Queue

abdulhuq811:
When using replicas for the workflow and the control plane with redis message queue, all the consumers consume and process the message. Adding support to enable exclusive consumer mode so that only one message consumer consumes and processes the message.


abdulhuq811:
@masci Can you help with reviewing this?

masci (approved):
LGTM!

PR #499, opened 2025-05-13
feat: expose nextjs UI from the apiserver

masci:


examples/quick_start/quick_start.yml
```diff
@@ -12,3 +12,9 @@ services:
       type: local
       name: src
     path: workflow:echo_workflow
+
+ui:
+  name: My Nextjs App
+  source:
+    type: local # could be git
```
logan-markewich, line 19: If the UI from create llama is "pullable", we could make `type: create-llama` or some other name work (tbh we can fork the create llama ui slightly to take advantage of more llama deploy features)

The one caveat is this will expect your workflow to have a particular input/output contract (similar for any prebuilt ui)

I wonder if there's a way to validate or communicate that input/output contract with deployed services

masci: This is super interesting, "source managers" are already pluggable and the code would be nicely encapsulated so I'm not concerned about the complexity of pulling in a create-llama app and validate it at deploy time. But to simplify things I think we should enforce matching the UI with the `default-service` workflow, probably a fair limitation.

logan-markewich: I agree!

examples/quick_start/quick_start.yml
```diff
@@ -3,12 +3,18 @@ name: QuickStart
...
-  dummy_workflow:
-    name: Dummy Workflow
+  echo_workflow:
+    name: Echo Workflow
     source:
       type: local
       name: src
     path: workflow:echo_workflow
+
+ui:
```
logan-markewich, line 16: We should add details for this section in the README for the example

examples/quick_start/ui/README.md
logan-markewich, line 1: Should we delete this?

logan-markewich (commented):
Is CSS working? 
image

masci:
@logan-markewich when you start the nextjs server from llamadeploy the static resources are mounted under the apiserver endpoint, so you have to access it via `http://localhost:4501/QuickStart/ui`

examples/quick_start/ui/app/page.tsx
```diff
@@ -0,0 +1,205 @@
...
+            <div className="mt-4">
+              <h3 className="text-lg font-medium mb-2">Result:</h3>
+              <pre className="p-3 bg-gray-100 dark:bg-gray-700 rounded-md overflow-auto text-sm">
+                {result}
+              </pre>
+            </div>
+          )}
+        </div>
+
+        <ol className="list-inside list-decimal text-sm/6 text-center sm:text-left font-[family-name:var(--font-geist-mono)]">
```
logan-markewich: Small nit, but we can remove some of this nextjs fluff

masci: Done 👍

logan-markewich (approved):

PR #502, opened 2025-05-20
feat: add `llamactl serve` command for local development

masci:
General concept:
- Added a new command `serve` to `llamactl` to boostrap a local LlamaDeploy instance running a specific deployment.
- When passing `--local` to `llamactl serve`, no code will be copied around, so changes to the local code will be picked up at the next restart (there's a plan to introduce a file watcher to make iterations even faster)
- Prior to this PR the concept of "local run" didn't exist, so I had to instruct several components to not try to sync the code listed in a deployment file. Several parts of the `apiserver` now takes a boolean flag called `local`.

For more specific comments to the code, see my review.

### DIFF
--- a/e2e_tests/apiserver/deployments/deployment_hitl.yml
+++ b/e2e_tests/apiserver/deployments/deployment_hitl.yml
@@ -11,4 +11,4 @@ services:
     source:
       type: local
       name: ./e2e_tests/apiserver/deployments/src
-    path: workflow_hitl:workflow
+    path: ./e2e_tests/apiserver/deployments/src/workflow_hitl:workflow
--- a/llama_deploy/apiserver/deployment.py
+++ b/llama_deploy/apiserver/deployment.py
@@ -293,10 +293,12 @@ class Deployment:
             self._set_environment_variables(service_config, destination)
 
             # Search for a workflow instance in the service path
-            pythonpath = (destination / source.name).resolve()
-            print(pythonpath)
+            module_path_str, workflow_name = service_config.path.split(":")
+            module_path = Path(module_path_str)
+            module_name = module_path.name
+            pythonpath = (destination / module_path.parent).resolve()
+            print(pythonpath, module_name)
             sys.path.append(str(pythonpath))
-            module_name, workflow_name = Path(service_config.path).name.split(":")
             module = importlib.import_module(module_name)
 
             workflow = getattr(module, workflow_name)
--- a/llama_deploy/apiserver/deployment.py
+++ b/llama_deploy/apiserver/deployment.py
@@ -338,6 +338,7 @@ class Deployment:
             for env_file in service_config.env_files:
                 # use dotenv to parse env_file
                 env_file_path = root / env_file if root else Path(env_file)
+                print("env_file_path", env_file_path)
                 env_vars.update(**dotenv_values(env_file_path))
 
         for k, v in env_vars.items():
--- a/llama_deploy/apiserver/deployment.py
+++ b/llama_deploy/apiserver/deployment.py
@@ -338,7 +338,6 @@ class Deployment:
             for env_file in service_config.env_files:
                 # use dotenv to parse env_file
                 env_file_path = root / env_file if root else Path(env_file)
-                print("env_file_path", env_file_path)
                 env_vars.update(**dotenv_values(env_file_path))
 
         for k, v in env_vars.items():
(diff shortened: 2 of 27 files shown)


e2e_tests/apiserver/deployments/deployment_hitl.yml
```diff
@@ -11,4 +11,4 @@ services:
     source:
       type: local
       name: ./e2e_tests/apiserver/deployments/src
-    path: workflow_hitl:workflow
+    path: ./e2e_tests/apiserver/deployments/src/workflow_hitl:workflow
```
masci, line 14: This is the right way to do this. Previously I was using `source.name` but that's counterintuitive for the users and difficult to manage for the maintainers. In a subsequent PR I'm going to clarify that when `source.type` is `local`, then `source.name` is not used.

llama_deploy/apiserver/deployment.py
```diff
@@ -59,15 +60,20 @@ class Deployment:
...
         Args:
             config: The configuration object defining this deployment
             root_path: The path on the filesystem used to store deployment data
+            local: Whether the deployment is local. If true, sources won't be synced
         """
+        self._local = local
         self._name = config.name
-        self._path = root_path / config.name
+        # If not local, isolate the deployment in a folder with the same name to avoid conflicts
+        self._path = root_path if local else root_path / config.name
```
masci, line 76: This replicates the previous behaviour, which is still the default since `local` defaults to `False`

llama_deploy/apiserver/deployment.py
```diff
@@ -282,20 +279,11 @@ def _load_services(self, config: DeploymentConfig) -> dict[str, WorkflowService]
                 raise ValueError(msg)
 
             # Sync the service source
-            destination = (self._path / service_id).resolve()
-
-            if destination.exists():
-                # FIXME: this could be managed at the source manager level, so that
```
masci, line 288: Implemented the FIXME, now this is taken care by the source managers

llama_deploy/apiserver/deployment_config_parser.py
```diff
@@ -75,7 +75,7 @@ class DeploymentConfig(BaseModel):
     message_queue: MessageQueueConfig | None = Field(None, alias="message-queue")
     default_service: str | None = Field(None, alias="default-service")
     services: dict[str, Service]
-    base_path: Path = Path()
+    base_path: Path | None = Field(None, alias="base-path")
```
masci, line 78: Moved the logic implementing the default value to the classmethod

llama_deploy/apiserver/source_managers/base.py
```diff
@@ -1,8 +1,18 @@
...
 from llama_deploy.apiserver.deployment_config_parser import DeploymentConfig
 
 
+class SyncPolicy(Enum):
+    """Define the sync behaviour in case the destination target exists."""
+
+    REPLACE = auto()
+    MERGE = auto()
+    SKIP = auto()
+    FAIL = auto()
```
masci, line 13: Existing managers use `SKIP` `MERGE` and `REPLACE` for now

llama_deploy/apiserver/server.py
```diff
@@ -42,8 +39,6 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None, Any]:
     apiserver_state.state("running")
     yield
 
-    t.close()
-    # Clean up deployments folder
-    if os.path.exists(manager._deployments_path.resolve()):
-        shutil.rmtree(manager._deployments_path.resolve())
```
masci, line 48: No need to clean up after exiting if the source managers do their job

llama_deploy/apiserver/deployment.py
```diff
@@ -347,6 +338,7 @@ def _set_environment_variables(
             for env_file in service_config.env_files:
                 # use dotenv to parse env_file
                 env_file_path = root / env_file if root else Path(env_file)
+                print("env_file_path", env_file_path)
```
logan-markewich: Debug print?

llama_deploy/apiserver/source_managers/base.py
logan-markewich: What is the FAIL policy for?

logan-markewich (commented):
Seems good to me honestly, not a scary change, just has to touch a lot of pieces

logan-markewich (approved):

adrianlyjak (approved):
LGTM. I think there's some things I want to test and think through for the deployment about private or workspace dependencies\*. I think it will likely work fine as long as the local environment for the workflow is the one with llamactl installed. 

\* Long term those private/workspace dependencies can hopefully be extracted, but they're a nice development convencience right now

PR #503, opened 2025-05-21
feat!: improve local source manager

masci:
There was a bit of confusion in the way local sources for a deployment are supposed to work, hopefully this should improve clarity. A few changes were made:
- Removed `base_path` from the deployment configuration. That's the path that's supposed to tell where the config file is, it doesn't make sense to keep it in the config file itself.
- Added `base_path` explicitly to the Deployment interface. Since a deployment configuration can be requested over the wire with no config file, in that case it's the user responsibility to define it
- Renamed `root_path` to `deployment_path` in the `Deployment` api. This should help clarify that `deployment_path` is where the workflows code is copied at deploy time.
- Changes to the `Deployment` class are breaking and this reflects on the Python client.
- The CLI will try to infer `base_path` whenever possible.

### DIFF
--- a/e2e_tests/apiserver/deployments/deployment2.yml
+++ b/e2e_tests/apiserver/deployments/deployment2.yml
@@ -11,5 +11,5 @@ services:
     host: localhost
     source:
       type: local
-      name: e2e_tests/apiserver/deployments/src
-    path: e2e_tests/apiserver/deployments/src:my_workflow
+      name: src
+    path: src:my_workflow
--- a/e2e_tests/apiserver/deployments/deployment2.yml
+++ b/e2e_tests/apiserver/deployments/deployment2.yml
@@ -10,6 +10,6 @@ services:
     port: 8002
     host: localhost
     source:
-      type: git
-      name: https://github.com/run-llama/llama_deploy.git
+      type: local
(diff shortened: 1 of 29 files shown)


e2e_tests/apiserver/deployments/deployment2.yml
```diff
@@ -10,6 +10,6 @@ services:
     port: 8002
     host: localhost
     source:
-      type: git
-      name: https://github.com/run-llama/llama_deploy.git
-    path: e2e_tests/apiserver/deployments/src:my_workflow
+      type: local
+      name: src
```
logan-markewich, line 14: so is name just a name, or is it still used as part of the path somehow?

If it has nothing to do with the path, can we give it a name other than `src` to make it less confusing? 😅

logan-markewich: I guess im confused how path and name work together

logan-markewich (approved):

e2e_tests/apiserver/deployments/deployment2.yml
logan-markewich: Clarified offline:
- name == where to find the code / root
- path == where in the code to import the workflow from

PR #511, opened 2025-05-25
feat: refactor message queue communication strategy

masci:
This PR rolls out a major refactoring of the way the Control Plane and the Workflow Service talk to the message queue.

This was the sequence of the operations before:
- ControlPlane calls a method on the MessageQueue to register itself as a consumer
- MessageQueue returns a callable that makes a POST request to the ControlPlane api at every message received
- ControlPlane must run that callable as a background task
- ControlPlane must run dedicated HTTP endpoints for the callable

This is the sequence of the operations after this PR:
- ControlPlane calls a method on the MessageQueue to register itself as a consumer
- MessageQueue returns a client object with a generic API but concrete implementation for the specific message queue type
- ControlPlane uses the client to consume messages

Advantages of this approach:
- No more HTTP calls involved in exchanging messages, much faster
- WorkflowService doesn't need any HTTP-server functionality anymore, required code is a fraction now
- Consumers can register as many times as they want, the MessageQueue doesn't care anymore
- Code for MessageQueue is much simpler now


llama_deploy/services/workflow.py
```diff
@@ -405,81 +371,48 @@ async def processing_loop(self) -> None:
...
-        port = self.config.internal_port or self.config.port
-        logger.info(f"Launching {self.service_name} server at {host}:{port}")
-
-        class CustomServer(uvicorn.Server):
-            def install_signal_handlers(self) -> None:
-                pass
-
-        cfg = uvicorn.Config(self._app, host=host, port=port)
-        server = CustomServer(cfg)
+        tasks = [
```
logan-markewich, line 403: So to clarify, the WorkflowService is now two pices
1. Iterating over messages and updating self._outstanding_calls
2. Actually processing those tasks

Very nice!

masci: Yes, before it was still two pieces I replaced the HTTP server (waiting for the message queue client to POST data) with a for loop consuming messages. It's not the best strategy, but if we want to support multiple message queues in an agnostic way, we don't have many alternatives to polling in a background task...

masci:
Merging before this gets too big to maintain in a branch

PR #517, opened 2025-05-27
llamactl init docs

logan-markewich:
Updating our docs and README to make `llamactl init` a first class citizen

Also took the time to remove docs that we don't want users seeing (low-level deployment pattern) and cleaning up some mkdocs warnings.


adrianlyjak (approved):
LGTM

masci (approved):
Noice! 🙏

PR #522, opened 2025-05-29
feat: imply --local when 'serve' is called passing a deployment file

masci:
The `--local` flag is hidden, and the behaviour of the resulting instance when not passed is confusing. Let's assume that if an user passes a deployment file like `llamactl serve /path/to/deployment.yaml`, a local execution is assumed.


logan-markewich (approved):
Makes sense!

I just realized I did not add this command to the docs yet (or mention it in the llamactl init command)

PR #524, opened 2025-05-30
Switch to base path, for nextjs ui router compatibility

adrianlyjak:


llama_deploy/apiserver/routers/ui.py
```diff
@@ -32,7 +32,7 @@ async def proxy(
     try:
         response = await client.request(
             method=request.method,
-            url=f"http://localhost:3000/{path}",
+            url=f"http://localhost:3000/ui/{deployment_name}/{path}",
```
masci, line 35: I think this is what I missed when I tried the `basePath` route without success

masci (approved):
Looks very good to me!

PR #525, opened 2025-05-31
Add streaming and web socket support to UI proxy

adrianlyjak:
This adds streaming support, both HTTP and websockets:
- next.js dev server works nicely now with hot reloading, and
- Can support SSE event style things now, when those are returned from the next.js backend

Also noticed basic example had issues with images not rendering, perhaps introduced by the `basePath` change. To workaround, this introduces some new environment variables. Unfortunately next.js just doesn't take `basePath` (or assetPrefix I think?) into account for images, so you manually need to prepend the path from some sort of configuration.

https://nextjs.org/docs/app/api-reference/config/next-config-js/basePath#images

image

### DIFF
--- a/examples/quick_start/ui/app/confetti/page.tsx
+++ b/examples/quick_start/ui/app/confetti/page.tsx
@@ -1,0 +1,64 @@
+'use client';
+
+import { useEffect } from 'react';
+import Image from 'next/image';
+import Link from 'next/link';
+import confetti from 'canvas-confetti';
+
+export default function LlamaPartyPage() {
+  // Trigger confetti on page load
+  useEffect(() => {
+    const timer = setTimeout(() => {
+      celebrate();
+    }, 500);
+
+    return () => clearTimeout(timer);
+  }, []);
+
+  const celebrate = () => {
+    // Fire confetti from both sides
+    confetti({
+      particleCount: 500,
+      spread: 90,
+      origin: { y: 0.6, x: 0.1 }
+    });
+    confetti({
+      particleCount: 500,
+      spread: 90,
+      origin: { y: 0.6, x: 0.9 }
+    });
+  };
+
+  return (
+    <div className="flex flex-col min-h-screen p-8 items-center justify-center font-[family-name:var(--font-geist-sans)]">
+      <Image
+        src={`${process.env.NEXT_PUBLIC_BASE_PATH}/logo-dark-light.svg`}
+        alt="LlamaIndex logo"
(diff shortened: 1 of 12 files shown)


examples/quick_start/ui/app/confetti/page.tsx
```diff
@@ -0,0 +1,64 @@
+'use client';
```
adrianlyjak, line 1: added an alternate route to manually validate routing is working

examples/quick_start/ui/app/page.tsx
```diff
@@ -123,7 +133,7 @@ export default function Home() {
         >
           <Image
             aria-hidden
-            src="file.svg"
+            src={`${process.env.NEXT_PUBLIC_BASE_PATH}/file.svg`}
```
adrianlyjak, line 136: Unfortunately, next.js doesn't take basePath into consideration for images. They say as much in their documentation. Added a new environment variable to this example (with NEXT_PUBLIC_, such that it's shippable to the client)

tests/apiserver/routers/test_deployments.py
```diff
@@ -217,7 +221,7 @@ def test_send_event(http_client: TestClient, data_path: Path) -> None:
         mocked_manager.get_deployment.return_value = deployment
 
         serializer = JsonSerializer()
-        ev = HumanResponseEvent(response="test human response")
+        ev = SomeEvent(response="test human response")
```
adrianlyjak, line 224: There was a kinda breaking change here at https://github.com/run-llama/llama_index/commit/a410814, that was causing type checking to fail here, as the `response` field was removed. This was only on the build server for the mypy/linting checks. Build server tests passed. My local environment seemed to stick with the old version for linting, perhaps the build server isn't using the lock?

Either way, just used a different event to make this compatible with more versions

llama_deploy/apiserver/routers/ui.py
```diff
@@ -1,54 +1,196 @@
...
+@ui_router.api_route(
+    "/{deployment_name}",
+    methods=["GET", "POST", "PUT", "DELETE", "OPTIONS", "HEAD", "PATCH"],
+)
 async def proxy(
     request: Request,
     deployment_name: str,
     path: str | None = None,
-) -> Response:
+) -> StreamingResponse:
```
masci, line 133: 💯

masci (approved):
Thanks for the extra comments, those made the review easy peasy!

PR #531, opened 2025-06-04
feat: move ui path under deployments router

masci:
Rethink the URL path to the user interface for a deployment:
- Before: `/ui/my_deployment_name`
- After: `/deployments/my_deployment_name/ui`


adrianlyjak (approved):
nice

PR #536, opened 2025-06-10
Add rudimentary event model serialization

adrianlyjak:
Adds an additional field: `"event_model"` that contains meta information about the event. I added a single object field with embedded sub-fields to minimize overwriting existing fields, while keeping the event more or less backwards compatible.

I think human response events may be broken? It does not seem to be resuming the workflow. See video

the task+session id ergonomics is a little awkward. I was envisioning a process where rich agentic multi-turn chat experiences would be implemented as separate runs of a workflow, with the workflow context retained. However getting a serial history of the events appears non-trivial. They are pretty baked in, but I'm wondering about making the API's more expose a single id, and events get added to the current active task (and only one task is allowed to be active at a time).

Additionally, getting a serial log of input AND output events appears mostly unsupported from the existing `deployments/{id}/tasks/{task_id}/events?session_id={session_id}` streaming API
- write_event_to_stream events are sent
- Some sort of StopEvent event is sent, but it seems to be missing the result field
- Result can be queried separately from `deployments/{id}/tasks/{task_id}/result?session_id={session_id}`

I'm not clear whether the input or human response events are visible from any api after sending them, I don't think so.

See video for more or less how it works: https://www.loom.com/share/43423b9db94940a9a32293bb073321ae?sid=045c799a-825c-40bf-9302-18f0f7f74f24


examples/quick_start/src/workflow.py
```diff
@@ -1,16 +1,50 @@
 import asyncio
+from typing import Literal
```
marcusschiesser, line 2: @adrianlyjak i guess we don't want to merge this - it's just for testing HITL, right?

llama_deploy/types/core.py
```diff
@@ -104,114 +104,62 @@ class QueueMessage(BaseModel):
...
-        service_id (str):
-            The service ID that the task should be sent to.
-            If blank, the orchestrator decides.
-    """
-
-    input: str
-    task_id: str = Field(default_factory=generate_id)
-    session_id: str | None = None
-    service_id: str | None = None
+    input: str = Field(description="The task input.")
```
marcusschiesser, line 107: @adrianlyjak as far as i can see this just adds metadata to the pydantic models, correct?

llama_deploy/services/event_serialization.py
```diff
@@ -0,0 +1,38 @@
...
+    trait: Optional[Literal["start", "stop", "wait"]] = None
+    if isinstance(event, StartEvent):
+        trait = "start"
+    elif isinstance(event, StopEvent):
+        trait = "stop"
+    elif isinstance(event, InputRequiredEvent):
+        trait = "wait"
+    metadata = {
+        "type": event.__class__.__name__,
+        "trait": trait,
```
marcusschiesser, line 31: what is the trait used for?

adrianlyjak: The idea was to have some sense of whether this was a start/step/wait/human event

adrianlyjak: Not entirely sure if its necessary

llama_deploy/types/core.py
adrianlyjak: Yes, this was totally tangential. I was looking through the API docs and noticed the descriptions were a little awkward in the openapi as I was trying to understand the models. Probably better to throw this out to keep it simple.

examples/quick_start/src/workflow.py
adrianlyjak: Exactly

PR #538, opened 2025-06-12
feat: get nextjs port from deployment file

masci:
Let users control the TCP port used to run nextjs if a `ui` service was configured.

### DIFF
--- a/llama_deploy/apiserver/deployment_config_parser.py
+++ b/llama_deploy/apiserver/deployment_config_parser.py
@@ -96,7 +96,10 @@ class Service(BaseModel):
 
 
 class UIService(Service):
-    pass
+    port: int | None = Field(
+        default=3000,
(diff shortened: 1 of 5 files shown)


llama_deploy/apiserver/deployment_config_parser.py
```diff
@@ -96,7 +96,10 @@ def validate_fields(cls, data: Any) -> Any:
 
 
 class UIService(Service):
-    pass
+    port: int | None = Field(
+        default=3000,
+        description="The TCP port to use for the nextjs server",
+    )
```
masci, line 102: Overriding `port` from the base class `Service` to provide a default value should be supported by Pydantic but not 100% sure so please correct me if I'm wrong (it works tho)

logan-markewich: Yes this will override the default from the other base class 👍

PR #539, opened 2025-06-12
Install dependencies with UV

adrianlyjak:
Replaces installer from pip to uv, and adds support for syncing a uv project given a path to a pyproject.toml
- bootstraps uv with pip if unavailable in the current environment
- installs dependencies to the current python through a couple obscure uv configurations.` --prefix` for `uv pip install`, which will install to the given path instead of a venv, and `UV_PROJECT_ENVIRONMENT` for `uv sync`, which is the other version for managed projects. Also includes `--inexact` to not obliterate the other deploy level dependencies.
- I also noticed that the requirements.txt support seems to have broken? I think this was a thing. Added that in by resolving the path and prefixing requirements.txt files with `-r`
- Updates dockerfiles to copy in the current latest version of uv (0.7.13)


tests/apiserver/data/python_dependencies_kitchen_sink.yaml
```diff
@@ -0,0 +1,23 @@
...
+  type: simple
+  host: "127.0.0.1"
+  port: 8001
+
+default-service: myworkflow
+
+services:
+  myworkflow:
+    name: My Python Workflow
+    python-dependencies:
```
logan-markewich, line 16: Ngl it feels weird to have python-dependencies and uv-dependencies 

Does the user care if its pip or uv or poetry under the hood? Feels like everything should just be under pytnon-dependencies 

Curious on the motivation for splitting into its own field

adrianlyjak: I don't like it either, but do you have ideas for to declare the pyproject.toml within the python dependencies? Its just an array of string, I could do a thing like the requirements.txt, and just embed the path to the pyproject.toml. UV uses a PEP standard for declaring its dependencies, so that seems perhaps reasonable, however it might be misleading to folks who have poetry dependencies in their pyproject.toml. Those wouldn't be supported

masci: I was commenting the same in another module. Why don't we just `uv pip install` the strings in `python-dependencies`? With the exception of `requirements.txt` for which we have to pass `-r`, I think `uv pip install path/to/a/folder/containing/pyproject`, should work whether the project file is poetry or uv compatible no?

adrianlyjak: oh, that's way nicer! I wonder if it respects the lockfile? `uv pip install .` does seem to also install workspace members

adrianlyjak: Updated, removed mention of `uv` as well, just specified that it needs to be pep compliant for the pyproject.toml

adrianlyjak:
I see coveralls is unhappy. Unfortunately it's 500ing when I try to get details (its been really flaky?). I'll add some more tests

adrianlyjak:
ok, coveralls is happy now

examples/python_dependencies/deployment.yaml
```diff
@@ -0,0 +1,18 @@
...
+  echo_workflow:
+    name: Pretty Echo Workflow
+    source:
+      type: local
+      name: src
+    path: workflow:echo_workflow
+    python-dependencies:
+      - cowpy
+      - "."
+      - "requirements.txt"
```
masci, line 18: Nice!

llama_deploy/apiserver/deployment.py
```diff
@@ -304,6 +304,27 @@ def _load_services(self, config: DeploymentConfig) -> dict[str, WorkflowService]
 
         return workflow_services
 
+    @staticmethod
+    def _validate_path_is_safe(
+        path: str, source_root: Path, path_type: str = "path"
+    ) -> None:
+        """Validates that a path is within the source root to prevent path traversal attacks.
```
masci, line 311: ❤️

llama_deploy/apiserver/deployment.py
```diff
@@ -325,24 +346,75 @@ def _set_environment_variables(
...
+                    ]
+                )
+            except subprocess.CalledProcessError as e:
+                msg = f"Unable to install uv. Environment must include uv, or uv must be installed with pip: {e.stderr}"
+                raise DeploymentError(msg)
+
+        # Bit of an ugly hack, install to whatever python environment we're currently in
+        # Find the python bin path and get its parent dir, and install into whatever that
+        # python is. Hopefully we're in a container or a venv, otherwise this is installing to
+        # the system python
```
masci, line 400: I think it's safe to assume we're at least in a venv, the alternative is the global python env was already messed up by the user if code execution arrives here.

masci (approved):
Looks very good to me!

PR #540, opened 2025-06-16
Followup fixes for uv dependencies

adrianlyjak:
I had some trouble actually running code from installed uv dependencies. Part of the complexity here was attempting to install from a uv workspace, but there were also some bugs

1. small fix: I noticed that run_autodeploy was not actually updating the `rc_path` setting, 
2. for some reason, in `llamactl serve` environments, after `uv pip install`ing, new modules are available immediately. When testing in build container with package installed, a module reload was required
3. Local and git sync have inconsistent sync path behavior. Local sync includes the source path as a subdir of the destination, while git does not, For example:
  - copying from `./foo/bar` to `/tmp/deployments/QuickStart` will create `/tmp/deployments/QuickStart/foo/bar`. Now it just copies to `/tmp/deployments/QuickStart`
  - whereas a `https://github.com/foo/bar` will just copy the files to `/tmp/deployments/QuickStart/`
  
To account for this difference without introducing a change breaking the llama deploy deployment.yaml config, I added a `relative_path` method to the `SourceManager` as a workaround

I discovered this when attempting to sync a UI with a source of `/app/ui` (for workspace support reasons). The UI sync sort of has an assumption currently that the UI is only ever in `/ui`, and this local folder path copying was embedded into it. It's `source` path was somewhat ignored, and always attempted to run `pnpm` from a cwd of `target_folder / "ui"`. This now instead resolves the cwd of `pnpm` commands to the `target_folder / relative_path`

4. Finally, the UI file sync is currently overwriting the previously cloned python workflow code. This mostly works, as the code is already loaded into memory, so it doesn't seem to cause issues? However as a safeguard, I added the option to instead merge in the UI via a `sync_policy` (This is also a somewhat unstable way to get the UI pnpm workspace to work, but the sync policy option seems reasonable either way)

### DIFF
--- a/docker/run_autodeploy.py
+++ b/docker/run_autodeploy.py
@@ -45,8 +45,7 @@ def setup_repo(
         run_process(["git", "pull", "origin", "main"], cwd=str(dest_dir.absolute()))
 
 
-def copy_sources(work_dir: Path, deployment_file_path: Path) -> None:
-    app_folder = deployment_file_path.parent
+def copy_sources(work_dir: Path, app_folder: Path) -> None:
     for item in app_folder.iterdir():
         if item.is_dir():
             # For directories, use copytree with dirs_exist_ok=True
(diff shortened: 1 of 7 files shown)


masci (approved):
Good catch!

PR #541, opened 2025-06-17
Add an option to get raw serialized event in event stream

leehuwuj:
Use the default llamaindex's JsonSerializer to serialize whole event object and send to event stream instead of just event data.

+ Benefit:
We have the same format of format when receiving and sending event to external system (as we are also using it in sending API, so it makes sense to me to have the same format in receiving API).
+ Disadvantage:
The format doesn't look nice for external system with unnecessary fields like `__is_pydantic` or `qualified_name`. 
-> I think later on we can consider using a mapping to map the name with qualified name and store the mapping in deployment file or workflow context? Then we can get rid of the `__is_pydantic` and `qualified_name`

For backward compatibility, I added a new flag to the `/{deployment_name}/tasks/{task_id}/events` API that the user can choose whether they want to get raw event or not.


marcusschiesser (approved):

docs/docs/module_guides/llama_deploy/apiserver.json
leehuwuj, line 1: I copied this file from the apiserver's Swagger UI. It seems that we haven't updated it recently.

marcusschiesser: we need the latest version for the TS client

marcusschiesser (approved):

marcusschiesser (approved):

masci (approved):
LGTM!

PR #542, opened 2025-06-18
refact: bypass message queue and control plane from apiserver

masci:
Given the current usage of LlamaDeploy through the apiserver in a single docker container, it doesn't make sense to carry on the massive overhead of sending messages back and forth the message queue.
With this PR:
- message queue and the control plane are completely bypassed
- apiserver Manager and Deployment don't start anymore the corresponding services for control plane, message queue server (when Simple was used) and each workflow listed in the deployment file
- Workflow Context and handler instances are kept in-memory in the Deployment object


e2e_tests/apiserver/deployments/deployment_env_git.yml
```diff
@@ -8,7 +8,7 @@ services:
     name: Git Workflow
     source:
       type: git
-      name: https://github.com/run-llama/llama_deploy.git
+      name: https://github.com/run-llama/llama_deploy.git@massi/refact
```
masci, line 11: to be changed after this PR is merged

logan-markewich (commented):
This seems great! The code is much easier to follow

But two things I noticed
1. We kept all the message queue code/files/examples/docs -- did we want to delete that?
2. Does `llamactl init` still work? It might still ask for a message queue because the pydantic config models were not touched

masci:
> But two things I noticed
> 
>     1. We kept all the message queue code/files/examples/docs -- did we want to delete that?
> 
>     2. Does `llamactl init` still work? It might still ask for a message queue because the pydantic config models were not touched

This was my question for you actually 😅. Technically speaking the "low level" api still works, if somebody doesn't use the apiserver nothing really changes for them. But we need to make a call whether we want to double down on the apiserver making it the only way to use this project, and in that case I would remove all the code/docs/exampels to avoid confusion. Depending on this decision I would also amend the `init` command, and I would follow up with a dedicated PR.

logan-markewich:
(Discussed offline) -- We should likely remove all reference of message queues/control planes in an attempt to fully commit to the single-container apiserver+workflows approach

logan-markewich (approved):
This looks good, it would be good to tackle the removal/clean-up in another PR and then we can make a release?

PR #548, opened 2025-06-20
feat: Expose workflows as MCP tools

masci:
- Create a global `FastMCP` instance along with the apiserver and mount it under the `/mcp` endpoint
- When creating a deployment, add each service as tool to the mcp server


logan-markewich (approved):
Nice, this looks pretty easy to do!

I think rather than using context.log(), theres probably a way to make workflow events be push notifications (and also eventually supporting their HITL thing). But we can figure that out later

PR #549, opened 2025-06-20
Adds basic MCP integration

adrianlyjak:
This would be better to mount to each deployment at a sub-path, but this was just a quick hack as a demo. This shouldn't be merged, just putting it here as a reference.

It includes an example mcp server that acts as a knowledge base that can scrape data from the web, to act as a documentation helper.

In general, I found the event formatting and the meta data (descriptions) to be the more complex part. If we pursue something like this, it might be nice to add some mechanism to be able to have a format Event outputs to one more suitable for a model's consumption, rather than just dumping out the json.


llama_deploy/apiserver/app.py
```diff
@@ -15,6 +17,7 @@
 
 
 app = FastAPI(lifespan=lifespan)
+app.mount("/", mcp_app.sse_app())
```
logan-markewich, line 20: sse is deprecated on MCP, FYI

adrianlyjak: yeah, I know. It was working for a bit (added it to the lifespan with the async context manager), but then something went sideways, and it consistently stopped working. I stopped debugging. The vibes I'm getting are that SSE is deprecated, but streaming is not yet very stable 😭

PR #550, opened 2025-06-23
Support cloning branches

adrianlyjak:
Add support for cloning branches with `<url>@<branch-name>` syntax

This is somewhat duplicative of our internal work with kubernetes deployments, but nice to have in the meantime. Architecturally, I think maybe moving towards having a way to prepare a container from a git repo, (or at least the files for a generic container), outside of startup, might reduce the duplication and improve startup times

Manually tested on a branch with some tweaks to build from local llama deploy https://github.com/run-llama/llama_deploy/blob/adrian/test-autodeploy/test_autodeploy.sh


masci (approved):
We can maybe migrate this code into the library at some point, so we can test it and remove some duplication

PR #559, opened 2025-07-29
fix: allow plain http

adrianlyjak:
Want to be able to clone git repositories inside of a private network. HTTPS requirement makes this difficult


masci (approved):
LGTM, not sure why I put that guard there 🙈

PR #567, opened 2025-10-02
Fixes: Handle missing tasks, use session_id passed as part of Task Definition, and handle AgentOutput in task result

abdulhuq811:
## Description

When using the current version of Llama Deploy, ran into a few issues:

- If a task is submitted and handler does not have the Task ID yet, it throws a KeyError which results in the server not respecting the poll interval. The fix here is to catch the KeyError and raise a 404 error to allow the server to poll based on the poll interval.
- If session_id is passed in as part of the Task Definition, it is not used, instead it is treated as None and a new session is created. Resolution for this is to allow setting the session_id if passed in as part of the Task Definition, this allows for re-using session context.
- When the output type is AgentOutput, then the get_task_result endpoint fails as the result is not of type string. Resolve this by converting AgentOutput to string and using that in the result.

Added appropriate unit tests for the changes.


abdulhuq811:
@masci Can you help review these changes? Thanks!

PR #584, opened 2026-04-06
Deprecate llama-deploy in favor of workflows-py

adrianlyjak:
## Summary

- Replace README content with a deprecation notice pointing to [run-llama/workflows-py](https://github.com/run-llama/workflows-py)
- Add a `DeprecationWarning` on `import llama_deploy` so users are notified at runtime


examples/mcp/src/mcp/workflows.py
logan-markewich, line 1: deprecating and adding an example? 😁

